PR #26, opened 2026-04-27
spec: Module 22 — Community validation (expert ID queue)

ArtemioPadilla:
Spec completo para cola de validación comunitaria. Ver docs/specs/modules/22-community-validation.md

PR #28, opened 2026-04-27
feat: Module 22 — Community validation implementation

ArtemioPadilla:
## Summary

End-to-end implementation of the community-validation queue and research-grade auto-promotion, per the [Module 22 spec](https://github.com/ArtemioPadilla/rastrum/blob/main/docs/specs/modules/22-community-validation.md) (v1.3, merged in #26).

- Migration: 1 SQL view + 3 RLS policies + 3 triggers + 1 partial UNIQUE index + tie-handling guard inside `recompute_consensus()`. All idempotent. **Zero new tables.**
- UI: `ValidationQueueView` + `SuggestIdModal` + `ValidationDashboardView` + 4 route pages (en/es × queue/dashboard).
- Existing surfaces: research-grade 🔬 chip wired into `MyObservationsView` and `ExploreRecentView`. Explore dropdown + Profile menu gain links.
- i18n: new `validation` block in both `en.json` and `es.json` (~30 keys).

## Architecture (recap from spec)

Community votes are NOT a new table. They are additional rows in the existing `identifications` table with `validated_by = auth.uid()`. The existing `recompute_consensus()` function already handles weighted aggregation (3× for `is_expert AND kingdom = ANY(expert_taxa)`); we plug into it via PostgREST RPC after each suggestion insert. The UI calls `select public.recompute_consensus($obs_id);` and the DB does the rest.

## What you need to do after merge

1. `make db-apply` against the live database to bring the view + policies + triggers online.
2. `make db-verify` — confirm `validation_queue` and the three new identification policies + three new triggers exist.
3. Hard-refresh `rastrum.org` to pick up the new bundle.

The four routes will then be live at:
- `/es/explorar/validar/` ↔ `/en/explore/validate/`
- `/es/perfil/validar/` ↔ `/en/profile/validate/`

## Verification

- ✅ `npm run typecheck` clean
- ✅ `npm test` 372/372
- ✅ `npm run build` 84 pages (was 80, +4 routes)

## Test plan

- [ ] Apply migration → confirm view + 3 policies + 3 triggers exist
- [ ] Visit `/es/explorar/validar/` signed-in → CTAs render correctly
- [ ] Submit a suggestion against a non-own observation → row INSERTs, `recompute_consensus` runs, no RLS errors
- [ ] Try to suggest on own observation → blocked at policy + trigger layer
- [ ] Two distinct users converge on same `taxon_id` → row promotes to `is_research_grade=true` + activity_event fires
- [ ] Tie scenario (two taxa with equal weighted score) → no promotion
- [ ] Research-grade chip 🔬 appears in `/es/perfil/observaciones/` and `/es/explorar/recientes/` for the promoted row
- [ ] `/es/perfil/validar/` dashboard shows correct counts


ArtemioPadilla:
## Response to Copilot review (`70ea9d9`)

21 line-level comments. Mapping each to its resolution.

### 🔴 SQL — duplicate / phantom infrastructure (the worst class of issue)

I duplicated three existing triggers. My earlier audit grepped for the names I had INVENTED in v1.1 of the spec, not the names that actually exist in the schema. After this commit, the migration is **strictly additive**:

| Original comment | Resolution |
|---|---|
| `tg_check_validator_not_observer` duplicates `prevent_self_validation` (~:599) | **Removed.** Existing trigger covers anti-self-vote on identifications. |
| `tg_research_grade_min_confidence` duplicates `enforce_research_grade_quality` (~:621) | **Removed.** Existing trigger enforces the 0.4 floor. |
| `tg_research_grade_activity_event` duplicates `fire_research_grade_trigger` (~:998); also used `visibility='self'` vs existing `'public'` | **Removed.** Existing trigger fires the activity event with the correct broader visibility. |
| `recompute_consensus()` redefined twice in same file | **In-place CREATE OR REPLACE.** Single definition; the tie-handling guard is folded into the original. |
| `validation_queue` re-queues 0.45-conf research-grade rows; doesn't handle NULL is_research_grade | **Fixed.** Predicate now requires `COALESCE(is_research_grade, false) = false` AND adds the confidence-low / no-id branch. |
| `id_validator_insert` allows votes on any observation_id (UUID guess attack on drafts) | **Fixed.** Policy now requires `sync_status='synced'` AND `obscure_level IN ('none','0.1deg','0.2deg','5km')` on the target obs. |

### 🟠 i18n — hard-coded strings

| Where | Resolution |
|---|---|
| SuggestIdModal: `submitting`, `failedTpl`, `noTaxonMatch`, target header | All sourced from data-* attrs reading new i18n keys (added 5 keys: `validate_signin_btn`, `research_grade_title`, `submit_target_tpl`, `nav_validate_label`, `submission_received`). |
| ValidationQueueView: subtitle hard-coded | Now uses `validation.queue_subtitle` template with `{n}` substitution. |
| ValidationQueueView: "Sign in" button label was the prompt sentence | New `validation.validate_signin_btn` key for the button text; prompt text stays separate. |
| ValidationQueueView, MyObservationsView, ExploreRecentView: research-grade pill copy hard-coded EN/ES | All three now read from the `validation.research_grade` / `research_grade_title` keys via root `data-*` attrs. |
| Header dropdown: "Validate" / "Validar" hard-coded | Now reads `validation.nav_validate_label`. |
| Route pages: description used `dashboard_subtitle` for the queue page | Replaced with page-specific copy. |

### 🟠 Bug fixes

| Issue | Resolution |
|---|---|
| Insert hits the partial UNIQUE index on repeat votes (Confirm twice or Suggest after Suggest) | Changed `.insert()` → `.upsert({...}, { onConflict: 'observation_id,validated_by' })`. |
| Toast always says "promoted to research grade" even when no promotion happened | SuggestIdModal now re-fetches the primary identification after `recompute_consensus()` and dispatches `promoted: boolean` in the event detail. Queue view branches between `submission_received` and `research_grade_promoted` toasts. |
| ValidationDashboardView `reachedRG` always 0 (community votes are is_primary=false but consensus promotes is_primary=true row with validated_by=NULL) | Now does a two-step query: fetch user's suggestions → fetch the corresponding primary identifications by `observation_id` → count where the user's taxon_id matches and the primary is research-grade. |
| ValidationDashboardView `queuePath` hard-coded | Now uses `getLocalizedPath(lang, routes.exploreValidate[lang] + '/')`. |

### 🟢 Security

| Issue | Resolution |
|---|---|
| `escapeText()` didn't escape `&` (attribute injection on observer names) | Now escapes `&` first, then `<`, `>`, `"`, `'`. Applied universally. |
| Datalist `innerHTML` only escaped `"` | Replaced with DOM API: `replaceChildren(...rows.map(...createElement('option')))`. Browser handles escaping. |

### Verification

- ✅ `npm run typecheck` clean
- ✅ `npm test` 372/372
- ✅ `npm run build` 84 pages

Issue #6, opened 2026-04-25
bug(photo-id): species not auto-identified after photo upload


ArtemioPadilla:
## Bug — Photo ID Not Triggering

**Reported by:** Eugenio Padilla (onboarding 2026-04-25)
**Device:** Android (CFE TEIT)
**Location:** /es/observar — ObservationForm

## What happened
After uploading a photo, no species identification appears. The scientific name field stays empty. PlantNet + Claude Haiku cascade is not triggering on photo upload.

## Expected behavior
Within 5-10 seconds of uploading a photo, a species suggestion should appear in the identification field with a confidence score.

## Possible causes
- `identify` Edge Function not deployed to production
- ANTHROPIC_API_KEY or PLANTNET_API_KEY not set in Supabase secrets
- Photo upload to R2/Storage failing silently before identify is called
- Client-side identify trigger not firing on mobile

## Steps to reproduce
1. Open /es/observar on Android
2. Upload a field photo
3. Wait — no species suggestion appears

ArtemioPadilla:
**Addressed** but requires operator action to fully take effect.

What shipped (`identification-block`):
- Visible identification chip in the observation form with spinner during PlantNet call, result chip on success, manual override at all times.
- PlantNet key resolution chain: `window.__RASTRUM_PLANTNET_KEY__` → `PUBLIC_PLANTNET_KEY` build env → BYO key from `localStorage` → 'configure a key in Profile → Edit' hint with link.
- 15 s timeout via AbortController; failures surface a clear message instead of silent stall.
- Edge Function `identify` deployed and verified (the smoke test `curl /functions/v1/identify` returns 200).

To make first-time visitors get instant ID without any setup, the operator (you) needs to:
```bash
gh secret set PUBLIC_PLANTNET_KEY    # paste your project key
gh workflow run deploy.yml
```
This bakes the key into the static build so any visitor calls PlantNet directly. The 500/day quota is shared but enough for launch volume.

Until that secret is set, anonymous visitors see the 'configure a key' hint instead of an automatic suggestion. Any signed-in user with their own BYO key in `/profile/edit` already gets instant ID today.

Closing this ticket once `PUBLIC_PLANTNET_KEY` is set and verified.

ArtemioPadilla:
[...]

Issue #5, opened 2026-04-25
bug(gps): location not auto-filling on Android mobile


ArtemioPadilla:
## Bug — GPS Auto-fill

**Reported by:** Eugenio Padilla (onboarding 2026-04-25)
**Device:** Android (CFE TEIT)
**Location:** /es/observar — ObservationForm

## What happened
'Obteniendo tu ubicación…' spinner shows but GPS coordinates never populate. User has to tap 'Ingresar coordenadas manualmente'.

## Expected behavior
On page load, browser requests geolocation permission → coordinates auto-fill within 5-10 seconds.

## Possible causes
- `navigator.geolocation.getCurrentPosition()` called but permission denied silently on Android
- Timeout too short for GPS cold start
- HTTPS required for geolocation (check if site is served over HTTPS)
- Android Chrome may require explicit permission prompt

## Steps to reproduce
1. Open /es/observar on Android Chrome
2. Upload a photo
3. Watch location field — stays 'Obteniendo tu ubicación…' indefinitely

ArtemioPadilla:
**Partially addressed** by `gps-two-pass` (shipped v1.0).

What changed:
- Two-pass GPS: a fast coarse fix (low-accuracy, ~1–3 s) populates the form immediately, then a high-accuracy refinement (10–15 s) tightens it.
- Form now blocks submit on non-finite (NaN) coords, so a partial fix that fails late surfaces a clear error instead of silently breaking sync.
- Manual coordinate fallback button still works when geolocation is denied or unavailable.

What's NOT in the fix:
- Android Chrome can still silently deny permission if the user previously declined and never reset. That's a one-time browser setting (Chrome → Site settings → Location → Allow), not something the app can override.
- We don't yet show a 'permission denied — open browser settings' message; we just fall through to manual entry. That's a real UX gap; opening as a follow-up would help non-technical users.

Please **retest at https://rastrum.org/es/observar/** on the same Android device — if GPS still doesn't auto-fill, open Chrome → Site settings for rastrum.org → Location → check it's set to Allow. Reopen if the issue persists with permission granted.

ArtemioPadilla:
**Followup shipped** (commit `8b2ee01`).

`getLiveGPS()` now catches `GeolocationPositionError` and surfaces a specific localized message in amber depending on the failure mode:

[...]

Issue #3, opened 2026-04-25
ux(auth): OAuth consent screen shows raw Supabase domain instead of rastrum.artemiop.com


ArtemioPadilla:
## UX Issue — Onboarding Bug

**Reported by:** Eugenio Padilla (first user, onboarding session 2026-04-25)

## What happened
During Google OAuth login, the consent screen shows the raw Supabase project URL instead of the Rastrum domain. Non-technical users may think it's phishing.

**Seen:** reppvlqejgoqvitturxp.supabase.co
**Expected:** rastrum.artemiop.com (or 'Rastrum' app name)

## Fix
- Supabase Dashboard → Project Settings → Auth → Custom Domains → set auth.rastrum.artemiop.com
- OR update Google Cloud Console OAuth app name to display 'Rastrum' prominently

ArtemioPadilla:
**Deferred — requires Supabase Pro plan ($25/mo).**

Custom auth domain (`auth.rastrum.org` instead of `<project>.supabase.co`) is a Supabase Pro feature. Rastrum is currently zero-cost target (free tier across Supabase, Cloudflare R2, GitHub Pages), so this stays deferred until either:

1. The project switches to a funded operating model (grant, sponsorship, donations), OR
2. Supabase adds custom domains to the free tier.

The OAuth flow itself works correctly today — Google + GitHub both redirect successfully and `session.user.user_metadata` populates the avatar + display name via the new `handle_new_user()` trigger. The only UX cost is the consent screen showing `reppvlqejgoqvitturxp.supabase.co` instead of a branded URL.

Tracked under roadmap item `oauth-custom-domain` in v1.0 with explicit blocker: 'Supabase Pro plan ($25/mo) — deferred for zero-cost target'.

Leaving open as a known-deferred so it surfaces if/when the funding situation changes.

ArtemioPadilla:
**Correction to my earlier comment** — Supabase Pro isn't needed for the branding fix; the OAuth consent screen logo + app name are controlled at the **OAuth provider** (Google Cloud Console / GitHub Developer Settings), not in Supabase. Only the *URL* (auth.rastrum.org vs the raw Supabase URL) requires Supabase Pro.

Free fix that mitigates 90% of the UX issue:

**Google:**
1. https://console.cloud.google.com/apis/credentials/consent → pick your project → Edit App
2. App name: `Rastrum`
[...]

Issue #2, opened 2026-04-25
bug(mobile): header navigation renders twice on mobile


ArtemioPadilla:
## Bug Report

**Reported by:** Eugenio Padilla (first real user, field biologist)
**Device:** Android (CFE TEIT)
**URL:** rastrum.artemiop.com/es/
**Date:** 2026-04-25

## What happened
The header/navigation renders twice on mobile. The full nav (Inicio, Observar, Mapa, Acerca de + Docs dropdown) appears duplicated on the same page.

## Screenshot
First user session screenshot shows nav repeated twice — once with the dropdown open, then again below the auth buttons.

## Expected behavior
Navigation appears once.

## Impact
High — affects all mobile users on first visit. Confusing UX.

## Likely cause
Possible double-render in `Header.astro` or layout wrapping the header twice on mobile viewport.

PR #36, opened 2026-04-28
feat(karma): Module 23 Phase 1 — karma + per-taxon expertise + rarity-weighted rewards

ArtemioPadilla:
## Summary

Implements **Module 23** (karma + per-taxon expertise + rarity-weighted rewards) — Phase 1 only. Realizes the v1.2 future-work line in module 22 (community validation).

- **Karma engine in-database**: `award_karma()` runs inside `recompute_consensus` when an observation crosses into research-grade. No new Edge Functions in v1.
- **Per-taxon expertise** at any rank via `user_expertise` + `taxa.ancestor_path` (uuid[] GIN-indexed array). Most-specific-ancestor wins.
- **Rarity-weighted rewards**: percentile-bucket multipliers (1.0× common → 5.0× ultra-rare) refreshed nightly via pg_cron.
- **Asymmetric +5/−2 with grace** (30 days OR 20 votes): losses only fire when a deeper-expertise voter wins consensus. Penalty caps at −4 even on rare species.
- **UI surfaces**: pre-vote microcopy in suggest modal, profile karma section with top-5 expertise, locale-paired Pokédex (`/profile/dex` + `/perfil/dex`).

Full design rationale: `docs/superpowers/specs/2026-04-27-karma-expertise-rarity-design.md`
Implementation plan: `docs/superpowers/plans/2026-04-27-karma-expertise-rarity-phase-1.md`

## What's NOT in this PR (Phase 2/3 — separate plans)

- Win toast confetti, weekly digest email, regional leaderboards, "First in Rastrum" badge, discovery-moment interstitials.
- IUCN/NOM-059 conservation-status bonuses (requires populating `taxa.iucn_category`, `taxa.nom059_status`, `taxa.is_endemic_mexico` from external sources).
- Tunable magnitudes via admin-editable `karma_config` table.

## Backwards compatibility

`users.is_expert` and `users.expert_taxa` remain readable. A migration shim hydrates `user_expertise` from these columns at apply time. The columns will be removed one release cycle later.

## Performance budget

Per consensus event: ~10–20 ms added (rarity lookup + ancestor_path overlap + ledger insert + user/expertise update). Storage at 10k-user scale: ~500 MB/year. Comfortable on Supabase free tier.

## Pre-merge operator actions

After merging, run in order:

```bash
make db-apply                                      # applies all idempotent SQL
make db-verify                                     # confirms tables, RLS, triggers
make db-cron-schedule                              # installs refresh-taxon-rarity-nightly
psql "$SUPABASE_DB_URL" -c "SELECT public.refresh_taxon_rarity();"  # seed buckets immediately
```

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run test` — 388/388 pass (8 new karma helpers tests)
- [x] `npm run build` — 86 pages, 0 errors, both `/en/profile/dex/` and `/es/perfil/dex/` present
- [ ] Operator: `make db-apply` succeeds idempotently (re-run safe)
- [ ] Operator: `make db-verify` shows the three new tables (`user_expertise`, `taxon_rarity`, `karma_events`) with RLS enabled
- [ ] Operator: `psql -c "SELECT public.refresh_taxon_rarity();"` returns void with no errors
- [ ] Operator: open `/en/profile/` while signed in — karma section renders, weekly delta shows 0 (no events yet)
- [ ] Operator: open `/share/obs/?id=<some-other-user's-obs>` while signed in — Suggest modal shows pre-vote microcopy with rarity stars

## Commits

18 commits — 12 plan tasks + 6 review-driven fixes (BEFORE-trigger semantics, NULL-rank loss bug, RLS opt-in gate, escAttr sanitization, etc.). Each task went through implementer → spec reviewer → code quality reviewer in subagent-driven development per `superpowers:subagent-driven-development`.


ArtemioPadilla:
@copilot review

ArtemioPadilla:
## Review — PR #36: feat(karma) Module 23 Phase 1

+3352 / -6 across 21 files. Large PR — reviewing the pieces that matter most: karma engine, profile integration, command palette, Pokédex, settings, and the new CI workflow.

---

### ✅ What's solid

**karma.ts + karma.test.ts** — Clean helpers, good test coverage on `rarityTier` and `formatDelta`. The rarity buckets (0.05 / 0.2 / 0.6) map well to the SQL percentiles.

**SuggestIdModal.astro (PR-36 version)** — Best version of the modal. The `renderDatalist` refactor to `document.createElement('option')` instead of `innerHTML` is the right call — browser handles escaping, avoids XSS from malformed `scientific_name` values. Keep this approach.

**ValidationQueueView** — `escapeText` helper is consistent and the CTA wiring via `window.__rastrumOpenSuggestId` is pragmatic for Astro's island model.

**db-apply.yml** — Good workflow. The `cancel-in-progress: false` on concurrency is critical — correct call. Idempotent SQL assumption is documented clearly. One note below.

**CommandPalette** — Solid a11y: `role="combobox"`, `aria-activedescendant`, `aria-selected`, keyboard nav (↑↓ Enter Esc). The debounce at 80ms is appropriate.

---

### ⚠️ Issues to address

**1. ProfileView — Stats row regressed (watchlist count is an extra DB call)**

```ts
// ProfileView.astro — current
const { count: watchlistCount } = await supabase
  .from('watchlist_entries')
  .select('*', { count: 'exact', head: true })
  .eq('user_id', user.id);
```

This fires after the profile already loaded — adds a waterfall. Consider rolling it into `stats_json` on the server (like `total_observations`) or at least batching it in `Promise.all` alongside the badges query and the recent observations fetches. Right now it's a third sequential roundtrip.

**2. hydrateKarma() — three extra queries on top of the main profile load**

Profile page now does: auth → profile → watchlist → observations → identifications + media → karma + karma_events + expertise = 7+ DB calls sequentially. That's going to feel slow on first load. Suggestion: move karma_total + weekly delta into a Postgres view or a column on `users` (it already has `karma_total`), and lazy-load the expertise list only if the karma section is in the viewport.

**3. db-apply.yml — the `git diff` for detecting changed files can be empty on force-push**

```yaml
changed=
```

If `github.event.before` is `0000000000000000000000000000000000000000` (first push to branch or force-push), this diff is empty and both `schema` and `cron` come out `false`, silently skipping the apply. Add a guard:

```yaml
if [ "${{ github.event.before }}" = "0000000000000000000000000000000000000000" ]; then
  echo "schema=true" >> "$GITHUB_OUTPUT"
  echo "cron=true"   >> "$GITHUB_OUTPUT"
  exit 0
fi
```

**4. PokedexView — missing empty state for users with zero synced observations**

The view renders a skeleton + grid but if `obsRows` is empty after filtering by `sync_status = 'synced'` there's no fallback message. Minor but visible to new users.

**5. ProfileView — two `escAttr` functions defined**

There's one in the main `<script>` block and another inside `hydrateKarma()`. Extract to a shared module or at least a single top-level definition.

**6. Settings page — the `tokensPath` construction is fragile**

```ts
const tokensPath = getLocalizedPath(lang, routes.profileExpertApply[...].replace('expert-apply', '') + 'tokens/');
```

String-replacing a route constant is a code smell — if the route key ever changes this silently breaks. Add `profileTokens` to the routes object explicitly.

---

### 🟡 Minor notes

- **CommandPalette labels fallback** — `LABELS[LANG] ?? LABELS['en']` is good but `LANG` comes from `document.documentElement.lang` which could be anything (e.g. 'en-US'). Consider `LANG.startsWith('es') ? 'es' : 'en'` to be safe.
- **PokedexView rarity badge** — the tier colors are duplicated from `karma.ts`'s `rarityTier`. Consider exporting a `tierBadgeClass` helper from `karma.ts` so the mapping lives in one place.
- **karma_events weekly delta** — the 7-day window uses `Date.now() - 7 * 86400000`. This is client-side time which could be skewed. Fine for display purposes but worth a comment so future readers don't expect server-side precision.

---

### Summary

The karma engine itself is clean and well-tested. Main concerns are the profile page becoming a waterfall of DB calls and a silent failure edge case in the CI workflow. Both are fixable without touching the schema. Everything else is polishable post-merge.

LGTM with those addressed 🦅

ArtemioPadilla:
## Review — PR #36: feat(karma) Module 23 Phase 1

+3352 / -6 across 21 files. Large PR — reviewing the pieces that matter most: karma engine, profile integration, command palette, Pokédex, settings, and the new CI workflow.

---

### ✅ What's solid

**karma.ts + karma.test.ts** — Clean helpers, good test coverage on `rarityTier` and `formatDelta`. The rarity buckets (0.05 / 0.2 / 0.6) map well to the SQL percentiles.

**SuggestIdModal.astro (PR-36 version)** — Best version of the modal. The `renderDatalist` refactor to `document.createElement('option')` instead of `innerHTML` is the right call — browser handles escaping, avoids XSS from malformed `scientific_name` values. Keep this approach.

**ValidationQueueView** — `escapeText` helper is consistent and the CTA wiring via `window.__rastrumOpenSuggestId` is pragmatic for Astro's island model.

**db-apply.yml** — Good workflow. The `cancel-in-progress: false` on concurrency is critical — correct call. Idempotent SQL assumption is documented clearly. One issue noted below.

**CommandPalette** — Solid a11y: `role="combobox"`, `aria-activedescendant`, `aria-selected`, keyboard nav (↑↓ Enter Esc). Debounce at 80ms is appropriate.

---

### ⚠️ Issues to address

**1. ProfileView — watchlist count is a sequential DB call**

```ts
// fires after profile load — adds a waterfall
const { count: watchlistCount } = await supabase
  .from('watchlist_entries')
  .select('*', { count: 'exact', head: true })
  .eq('user_id', user.id);
```

Roll this into `stats_json` server-side (like `total_observations`) or batch it in a `Promise.all` alongside the observations + media fetches. Right now it's a third sequential roundtrip when there are already 7+ DB calls on the profile page.

**2. hydrateKarma() — three extra queries on top of the main profile load**

The flow is now: auth → profile → watchlist → observations → identifications + media → karma + karma_events + expertise = 7+ roundtrips. Suggestion: lazy-load the expertise list only when the karma section enters the viewport, and rely on the existing `users.karma_total` column for the header number (it's already there — no extra query needed).

**3. db-apply.yml — silent skip on force-push / first push to branch**

```yaml
changed=$(git diff --name-only "${{ github.event.before }}" "${{ github.sha }}" || true)
```

When `github.event.before` is `0000000000000000000000000000000000000000` (force-push or first push to a branch), the diff is empty and both `schema=false` / `cron=false` — the apply is silently skipped. Add a guard at the top of the step:

```yaml
if [ "${{ github.event.before }}" = "0000000000000000000000000000000000000000" ]; then
  echo "schema=true" >> "$GITHUB_OUTPUT"
  echo "cron=true"   >> "$GITHUB_OUTPUT"
  exit 0
fi
```

**4. PokedexView — no empty state for users with zero synced observations**

If `obsRows` comes back empty (new user, all drafts), the skeleton hides but nothing else shows. Add a fallback message/CTA similar to what ProfileView does with `#no-observations`.

**5. ProfileView — duplicate `escAttr` definition**

One in the main `<script>` block, another inside `hydrateKarma()`. Extract to a shared util or define once at the top level of the component.

**6. Settings page — fragile `tokensPath` construction**

```ts
const tokensPath = getLocalizedPath(lang, routes.profileExpertApply[...].replace('expert-apply', '') + 'tokens/');
```

String-replacing a route constant is a code smell — if the key changes this silently breaks. Add `profileTokens` to the routes object explicitly.

---

### 🟡 Minor notes

- **CommandPalette lang detection** — `document.documentElement.lang` can return `'en-US'`; consider `LANG.startsWith('es') ? 'es' : 'en'` to be safe.
- **Rarity badge colors duplicated** — `PokedexView` and `ProfileView` both inline tier → color mappings that already live in `karma.ts`. Consider exporting a `tierBadgeClass(tier)` helper from `karma.ts`.
- **7-day delta uses client time** — fine for display, just worth a comment so future readers don't assume server-side precision.

---

### Summary

Karma engine is clean and well-tested. Main blockers are the profile page waterfall and the CI silent-skip edge case — both fixable without touching the schema. Everything else is post-merge polish. LGTM on the karma core 🦅

ArtemioPadilla:
Triaged the review feedback in `fc0a077`:

**Addressed:**
- ✅ **#3 db-apply.yml silent skip on first push / force-push.** Added explicit guard for `github.event.before == 0000...` case, defensively replays both SQL files (both are idempotent).
- ✅ **#5 Duplicate `escAttr`.** Hoisted to `src/lib/karma.ts` with three Vitest cases. Removed from both `ProfileView.astro` script blocks and `PokedexView.astro`.

**Deferred (post-merge polish, tracked for v1.1):**
- **#2 `hydrateKarma()` query waterfall.** ~100ms of additional latency below the fold. Lazy-load via IntersectionObserver is a clean v1.1 improvement.

**Out of scope (not in this PR):**
- **#1 Watchlist sequential roundtrip** — pre-existing in `ProfileView.astro` since profile streak landed.
- **#6 Settings `tokensPath` string-replace** — landed in PR #33 (Account hub / Settings shell).
- Minor: CommandPalette lang detection (PR #34), rarity badge color helper (no actual duplication exists), client-time delta (cosmetic).

Verified post-fix: typecheck clean, **11/11** karma unit tests pass (8 original + 3 escAttr), 86 pages build clean. Ready for re-review.

ArtemioPadilla:
## Module 23 Phase 1 — two-week health check

### Setup

A `workflow_dispatch`-only audit workflow was added in draft PR #1038
(`chore/karma-phase-1-audit`). It runs five diagnostic queries against prod
via `SUPABASE_DB_URL` and prints formatted tables to the action log. Because
`gh` CLI is unavailable in this CI context the dispatch couldn't be fired
automatically; see issue #1039 for the copy-pasteable queries and instructions
to fire it.

### Five health checks (schema confirmed ✓)

The module 23 schema was verified against the current
`docs/specs/infra/supabase-schema.sql`:

| # | Query | Tables/columns confirmed |
|---|---|---|
| (a) | Weekly `consensus_win`/`consensus_loss` counts | `karma_events.reason`, `created_at` ✓ |
| (b) | Avg `delta` per `rarity_bucket` on wins | `karma_events.rarity_bucket`, `delta` ✓ |
| (c) | Grace-period cohort split | `users.grace_until` (backfilled at apply time) ✓ |
| (d) | Top-5 karma users + top-5 expertise taxa | `users.karma_total`, `user_expertise.taxon_id` ✓ |
| (e) | Users with `karma_total < 0` | `users.karma_total` ✓ |

### What a healthy result looks like

- **(a)** Non-empty rows for both win/loss — confirms `recompute_consensus` is
  calling `award_karma()` on resolved identifications.
- **(b)** A clear gradient: `rarity_bucket 4–5` should show `avg_delta ≥ 2×`
  bucket 1, matching the `+5 × rarity_multiplier` formula. Flat buckets mean
  the `taxon_rarity` nightly cron hasn't seeded data or the lookup is
  returning `NULL`.
- **(c)** Most users should still be `in_grace` two weeks after launch (
  `grace_until = created_at + 30 days`). A small `exited_grace` pool of
  early adopters is normal. If both columns are zero the backfill didn't run.
- **(d)** Top karma user should be < 10× the median. A runaway outlier
  suggests a feedback loop (e.g. the same observation being re-evaluated).
- **(e)** Zero or low single digits is healthy. Double-digit negatives suggest
  the `−2 × min(rarity_multiplier, 2)` loss term is too aggressive for the
  current consensus volume.

### Phase 2 readiness call

**Hold 1–2 more weeks before starting Phase 2** (weekly digest + karma-gated
features). The reasons are structural, not a problem:

1. The grace-period cohort transition is still in progress — most users won't
   feel the loss penalty until day 30+. Calibration data for loss is thin
   until that cohort exits.
2. `taxon_rarity` multipliers are only as good as `refresh_taxon_rarity()`
   having enough observation history. Two weeks is on the edge; four weeks
   gives a full month of percentile data before Phase 2 UI surfaces
   `rarity_bucket` to users.
3. No blockers in the schema or logic — all five tables and the `award_karma()`
   function are confirmed present and correctly RLS-gated.

**Fire the queries in #1039 or via PR #1038 and post results here. If (a)
returns rows and (b) shows a gradient, the engine is healthy and Phase 2
can start.**

https://claude.ai/code/session_016xiemkRzWDPWdLXAuNCRd5

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #4, opened 2026-04-25
bug(i18n): typo 'Selecciona hábita' missing final t in observation form


ArtemioPadilla:
## Bug — Typo

**Reported by:** Eugenio Padilla (onboarding session 2026-04-25)
**Location:** /es/observar — ObservationForm, habitat dropdown placeholder

## What happened
Habitat dropdown placeholder reads 'Selecciona hábita' instead of 'Selecciona hábitat'.

## Fix
Find and fix in `src/i18n/es.json` or directly in `ObservationForm.astro`.

ArtemioPadilla:
Already correct in i18n/es.json: 'Selecciona hábitat (opcional)'. The typo was likely in an older build. Closing.

PR #42, opened 2026-04-28
Admin console (PR1 foundation): schema + chrome + Edge Function + 3 tabs

ArtemioPadilla:
## Summary

Foundation PR for the admin / moderator / expert console (Module 24).

- **Schema:** `user_roles` join table + `audit_op` enum + `admin_audit` table + `has_role(uid, role)` SECURITY DEFINER helper + sync trigger keeping `users.is_expert` / `credentialed_researcher` cached
- **Chrome:** new `'console'` mode in `chrome-mode.ts`, `ConsoleLayout` + `ConsoleSidebar`, role pills, conditional Console pill in `Header.astro` + `MobileDrawer.astro` (only visible when user holds a role)
- **Edge Function:** `supabase/functions/admin/` dispatcher with shared auth/audit helpers + 3 handlers (`role.grant`, `role.revoke`, `sensitive_read.user_audit`); writes are atomic with the `admin_audit` insert
- **3 working tabs:** Overview (empty-state shell), Experts (moved from `/profile/admin/experts/` with old path 308-redirected), Audit log (basic browse + JSONB diff drill-down)
- **Single source of truth:** `console-tabs.ts` declares all 25 tabs; sidebar, role pills, and routes are pure projections
- **EN/ES parity:** every route slug paired, all console.* i18n keys added to both `en.json` and `es.json`
- **Docs:** module spec 24, runbooks (bootstrap, role-model, audit, ops), user-facing console doc page, `CLAUDE.md` "Console / privileged surfaces" section, `architecture.md` flow diagram, roadmap entry in `progress.json` + `tasks.json`

PR2 (Users tab + Credentials + Expert console) and PR3 (Sync/API/Cron + Moderator console) get their own plans authored when each is next-priority.

Design rationale: [`docs/superpowers/specs/2026-04-27-admin-console-design.md`](docs/superpowers/specs/2026-04-27-admin-console-design.md)
Implementation plan: [`docs/superpowers/plans/2026-04-27-admin-console-pr1-foundation.md`](docs/superpowers/plans/2026-04-27-admin-console-pr1-foundation.md)

## Bootstrap (run after merge)

The schema migration auto-applies on merge via `.github/workflows/db-apply.yml`. Then to grant the operator (artemiopadilla@gmail.com) the first admin role:

\`\`\`bash
make db-psql -- -c "SELECT id FROM public.users WHERE email = 'artemiopadilla@gmail.com';"
make db-psql -- -c "INSERT INTO public.user_roles (user_id, role, granted_by, notes) VALUES ('<uuid-from-above>', 'admin', NULL, 'bootstrap') ON CONFLICT (user_id, role) DO NOTHING;"
make db-psql -- -c "SELECT public.has_role('<uuid-from-above>', 'admin');"  # expect: t
\`\`\`

After that, sign in to rastrum.org — the Console pill appears in the header.

The \`admin\` Edge Function deploy is gated to your manual trigger:
\`\`\`bash
gh workflow run deploy-functions.yml --ref main -f function=admin
\`\`\`

Full procedure: [\`docs/runbooks/admin-bootstrap.md\`](docs/runbooks/admin-bootstrap.md).

## Test plan

- [x] \`npm run typecheck\` — clean
- [x] \`npm run test\` — 411/411 pass (35 suites; 3 new test files: chrome-mode-console, console-tabs, admin-client)
- [x] \`npm run build\` — 94 pages, clean
- [x] \`npm run test:e2e -- tests/e2e/console-smoke.spec.ts\` — 10/10 pass on chromium + mobile-chrome
- [ ] After merge: schema applies via CI, run bootstrap SQL above
- [ ] After merge: deploy admin Edge Function via \`gh workflow run\`
- [ ] After bootstrap: sign in, verify Console pill appears, navigate to /en/console/, /en/console/experts/, /en/console/audit/


Pamela-ruiz9 (approved):
## Review — PR #42: Admin Console Foundation (Módulo 24)

**Veredicto: ✅ LGTM con comentarios menores**

Sólido. El diseño de fondo está bien pensado y el PR llega mucho más completo de lo que titula. Aquí va el breakdown:

---

### ✅ Lo que está bien

**Schema:**
- `has_role()` como SECURITY DEFINER con `SET search_path = public` — correcto, sin inyección de path
- El partial index `WHERE revoked_at IS NULL` es el corte adecuado; future-dated revocations se manejan en query time
- `admin_audit_no_client_write` como política explícita (no solo ausencia de policy) es la forma correcta de hacerlo
- Atomic write + audit en la Edge Function — bien, no hay ventana entre la acción y el log
- Sync trigger solo en `INSERT OR UPDATE OF revoked_at OR DELETE` — no dispara en cambios irrelevantes

**Edge Function:**
- `reason` mínimo 5 chars enforced en el dispatcher, no en cada handler — DRY correcto
- `verifyJwtAndLoadRoles` usa `getUser(jwt)` para verificar con Supabase antes de cargar roles — no confía solo en el JWT payload
- `requireRole` tira `HttpError(403)` en lugar de un genérico 500 — errores bien tipados

**Arquitectura:**
- `console-tabs.ts` como single source of truth para sidebar, rutas y role pills — decisión correcta
- EN/ES parity completa desde PR1, no como deuda futura — bien
- 308 redirect para la ruta legacy `/profile/admin/experts/` — no deja URLs rotas

---

### ⚠️ Comentarios menores (no bloqueantes)

**1. `getUserRoles()` en el cliente — PRIMARY KEY en DB ya previene duplicados, pero vale un comentario en la función.**

**2. Future-dated revocation no cubre el partial index** — `has_role()` lo maneja en query time, correcto. Si el patrón se vuelve frecuente, considerar índice funcional `WHERE revoked_at IS NULL OR revoked_at > now()`. Por ahora OK.

**3. `CORS_HEADERS: Allow-Origin: *` en la Edge Function** — para producción vale acotar al origin de rastrum.org. Deuda de seguridad conocida, anotar en runbook.

**4. `sensitive_read.user_audit`** — confirmar que `target_id` en el audit row incluye el user_id sujeto para que la consulta quede trazada.

---

### 🔍 Fuera de scope (PR2+)
- Rate limiting en Edge Function (especialmente `role.grant`) — sin esto un admin comprometido puede escalar masivamente
- Notificación al usuario cuando su role cambia

---

**Tests:** 411/411 + 10/10 e2e es buena cobertura para foundation.

**Merge cuando quieras.** Bootstrap SQL post-merge es claro.

🌙 *Review por Nyx*

ArtemioPadilla:
## Review — PR #42: Admin Console PR1 Foundation

+5294 / -40 across 52 files. Large but well-scoped: schema, chrome, Edge Function dispatcher, 3 working tabs, 411 tests passing, 94-page build. Reviewing the pieces that carry the most risk.

---

### ✅ What's solid

**Schema design** — `user_roles` as a join table (not columns on `users`) is the correct call. The `has_role()` SECURITY DEFINER helper that RLS predicates can call without exposing `user_roles` directly is well-thought-out. `audit_op` as an enum rather than free text means the audit log is queryable and the values are validated at the DB level.

**Edge Function dispatcher pattern** — `HANDLERS[action]` registry with a Zod payload schema + `requiredRole` + `op` per entry is clean and extensible. Adding a new privileged action = one file + one registry entry. The mandatory `reason` field (min 5 chars enforced at the dispatcher) + atomic audit write is exactly right for a sensitive surface.

**`console-tabs.ts` as single source of truth** — sidebar, role pills, and routes all derived from one array. 25 tabs declared with `stub: true` for phases 2–4; "soon" badges render automatically. This is the right architecture for incremental delivery.

**ConsoleLayout** — separate chrome from the main Header/Footer, no public nav, role pills with color-coded rails (emerald/amber/sky), identity strip, exit link. Clean separation of concerns.

**Tests** — 3 new test files (chrome-mode-console, console-tabs, admin-client) covering the new surface. `console-smoke.spec.ts` covers the gate, ES locale, 308 redirects, and the unauth pill state.

**308 redirect** from `/profile/admin/experts/` → `/console/experts/` is the right way to deprecate the old path without breaking existing links.

---

### ⚠️ Issues to address

**1. `role-grant.ts` — `expires_at` is used as `revoked_at` on upsert**

```ts
// role-grant.ts line ~43
.upsert({
  ...
  revoked_at: payload.expires_at ?? null,  // ← this sets the row as already-revoked
}, { onConflict: 'user_id,role' });
```

An `expires_at` is a forward-looking expiry timestamp. Setting `revoked_at` to it immediately means `has_role()` returns `true` only while `now() < expires_at`, which is actually the intended behavior — but it's semantically confusing and a future reader will misread it. More importantly: if an admin grants a role *without* `expires_at` (the normal case), the upsert sets `revoked_at = null`, which correctly reactivates a previously revoked row. That part is fine. But if an admin runs `role.grant` on an already-active role to *extend* its expiry, the `onConflict` upsert will overwrite `granted_at` and `granted_by` — losing the original grant history. Consider using `INSERT ... ON CONFLICT DO UPDATE SET revoked_at = EXCLUDED.revoked_at` without touching `granted_at`/`granted_by` on conflict.

**2. `auth.ts` — two Supabase clients created on every request**

```ts
const userClient = createClient(SUPABASE_URL, ANON_KEY, { global: { headers: ... } });
const adminClient = createClient(SUPABASE_URL, SERVICE_KEY);
```

Both are created fresh per request with no connection pooling. For a low-traffic admin surface this is fine, but the `adminClient` here shares the name with the `admin` client created in `index.ts` — if the auth module is ever extracted or reused, the variable collision will cause a confusing bug. Rename one (e.g. `serviceClient` in `auth.ts`).

**3. `user_roles` RLS — no policies defined**

The schema enables RLS on `user_roles` but doesn't add any policies:
```sql
ALTER TABLE public.user_roles ENABLE ROW LEVEL SECURITY;
-- (no CREATE POLICY follows)
```
With RLS enabled and no policies, the table is effectively blocked for all `authenticated` reads — including the `getUserRoles()` call in `user-roles.ts` which uses the anon/user client. The `has_role()` function works because it's `SECURITY DEFINER`, but any direct Supabase client query to `user_roles` from the front end will return empty. Add at minimum:
```sql
-- Self-read: users can read their own roles (needed for Header pill)
CREATE POLICY "user_roles_self_read"
  ON public.user_roles FOR SELECT
  USING (user_id = auth.uid());
```
Admin-write is handled by the Edge Function via service role — no client-side write policy needed.

**4. `console-tabs.ts` — 15 admin route keys referenced, only 3 routes exist in `utils.ts`**

`CONSOLE_TABS` references route keys like `consoleUsers`, `consoleCredentials`, `consoleObservations`, etc. for stub tabs. `tabsForRoles()` calls `routes[tab.routeKey]` which would throw at runtime if any stub tab is accessed. Since the sidebar renders all tabs (including stubs), this means `ConsoleSidebar` will throw `undefined` on `routes[tab.routeKey][lang]` for any stub tab.

Need either: (a) add all 25 route keys to `utils.ts` pointing at the same stub URL (e.g. the overview page), or (b) add a guard in `ConsoleSidebar`:
```ts
const href = routes[tab.routeKey]?.[lang]
  ? getLocalizedPath(lang, routes[tab.routeKey][lang] + '/')
  : '#';
```

**5. `admin_audit` — `ip` column type is `inet` but the input is a string**

```ts
ip: req.headers.get('x-forwarded-for'),  // string like "1.2.3.4, 10.0.0.1"
```

`x-forwarded-for` can be a comma-separated list when behind multiple proxies (`"1.2.3.4, 10.0.0.1"`). Inserting that string into an `inet` column will fail with a Postgres type error. Extract just the first IP:
```ts
ip: req.headers.get('x-forwarded-for')?.split(',')[0].trim() ?? null,
```

---

### 🟡 Minor notes

- **`researcher` role** — declared in the enum and in the TypeScript type but has no tabs in `CONSOLE_TABS` and `rolePillsFor` excludes it from pills. If it's not a console role (just a profile badge), consider removing it from the enum or documenting explicitly that it has no console surface.
- **`admin_audit` has no RLS policies** — same pattern as `user_roles`. Service role writes, so it works, but a self-read policy for admins to browse their own audit history from the client would be needed for the Audit Log tab (currently it probably works only because `ConsoleAuditLogView` calls the `sensitive_read.user_audit` Edge Function via service role, not directly).
- **Bootstrap SQL in PR description** uses `make db-psql` which isn't defined in `Makefile` (or there is no `Makefile`). Should be `psql $SUPABASE_DB_URL` per the existing `db-apply.yml` pattern.

---

### Summary

Architecture is solid — dispatcher pattern, single-source tabs, separate chrome, atomic audit writes. Three issues need fixing before merge: the `user_roles` missing RLS policies (will break the Header pill for real users), the stub route keys that will throw at runtime in `ConsoleSidebar`, and the `x-forwarded-for` → `inet` type mismatch. The `role-grant` `expires_at`/`revoked_at` semantic is technically functional but should be clarified. Rest is post-merge polish.

LGTM on the architecture once those four are addressed 🦅

ArtemioPadilla:
Rebased onto current `main` (resolved i18n conflicts with module 25's `privacy.*` block — both blocks now coexist as siblings). Force-pushed.

Addressed all four substantive issues from your review (commit `3624a14`):

- **`role-grant.ts` upsert overwriting `granted_at`/`granted_by`** — split the path: existing row → `UPDATE` only `revoked_at`; missing row → `INSERT` with full audit columns. History preserved on extends.
- **`auth.ts` variable collision (`adminClient`)** — renamed to `serviceClient` to disambiguate from the `admin` service-role client in `index.ts`.
- **`x-forwarded-for` → `inet` type mismatch** — extract first IP before insert; the `inet` column doesn't accept comma-separated proxy chains.
- **Stub route keys + `user_roles` RLS** — verified both as already correct: all 22 console route keys exist in `routes` (commit `6460857`); `user_roles_admin_or_self_read` and `user_roles_no_self_write` policies are in the schema at lines 2115-2126. (My earlier fix commit `8797fc1` added them after the first code review caught the original migration shipping with no policies.)

Plus quick polish from Pamela's review:
- JSDoc on `getUserRoles()` noting the PK (user_id, role) prevents duplicates.
- Explicit comment in `console-tabs.ts` that `researcher` is intentionally console-less (data-access role, no pill / no tab set).

Deferred to PR2/3 polish (per Pamela's notes):
- Tighten `CORS_HEADERS: Allow-Origin: *` to rastrum.org for production.
- Rate limiting on the dispatcher (especially `role.grant`).
- Optional: `admin_audit` self-read RLS policy for client-side browse (Audit Log tab currently goes through Edge Function dispatcher with service role, so works without it — but adding it would let the tab use the JS client directly).

Verification: typecheck clean, **449/449** unit tests (up from 411 thanks to module 25's tests joining via rebase), 94-page build, GitGuardian green. Merge state: `CLEAN`.

Ready for re-review.

Pamela-ruiz9 (commented):
## Re-review — respuesta al comentario de ArtemIO

Leí el código antes de opinar. De los 4 puntos, 2 están equivocados, 1 es válido, 1 es una cuestión de diseño.

---

### ❌ Punto 1: "user_roles sin RLS policies" — **incorrecto**

El schema incluye:
```sql
DROP POLICY IF EXISTS user_roles_admin_or_self_read ON public.user_roles;
CREATE POLICY user_roles_admin_or_self_read ON public.user_roles
  FOR SELECT TO authenticated
  USING (public.has_role(auth.uid(), 'admin') OR user_id = auth.uid());
```

La policy `user_roles_admin_or_self_read` existe. `getUserRoles()` en el cliente hace `SELECT ... WHERE user_id = userId`, que cae exactamente en el `user_id = auth.uid()` del USING. El Console pill va a funcionar para usuarios con roles. Este punto no aplica.

---

### ❌ Punto 2: "Stub route keys → undefined en runtime" — **incorrecto**

`utils.ts` define los 22 route keys de consola: `consoleUsers`, `consoleCredentials`, `consoleObservations`, `consoleApi`, `consoleSync`, `consoleCron`, `consoleBadges`, `consoleTaxa`, `consoleKarma`, `consoleFlags`, `consoleFeatureFlags`, `consoleBioblitz`, `consoleModFlagQueue`, `consoleModComments`, `consoleModBans`, `consoleModDisputes`, `consoleExpertValidation`, `consoleExpertExpertise`, `consoleExpertOverrides`, `consoleExpertTaxonNotes` — todos presentes. Los stubs en `console-tabs.ts` referencian keys que sí existen. No hay `undefined` en runtime.

---

### ✅ Punto 3: `x-forwarded-for` → inet — **válido y bloqueante**

`req.headers.get("x-forwarded-for")` puede devolver `"1.2.3.4, 10.0.0.1"` (comma-separated list según RFC 7239). Postgres rechaza eso como `inet`. Fix de una línea en `audit.ts` o en el dispatcher:
```ts
ip: (req.headers.get("x-forwarded-for") ?? "").split(",")[0].trim() || null,
```
Esto sí es un bug real que causa inserción fallida bajo proxies/load balancers.

---

### 🟡 Punto 4: `role-grant` sobreescribe `granted_at`/`granted_by` — **ya resuelto en el código**

Miré el handler. El `existing` check hace exactamente lo correcto:
- Si el rol ya existe → `UPDATE` solo `revoked_at`, **preserva** `granted_at` y `granted_by`
- Si no existe → `INSERT` con nuevos timestamps

El historial original **no** se pierde. Este punto está resuelto.

---

### Veredicto actualizado

El único bug real es el del inet (punto 3). Los demás no aplican al código actual. Recomiendo fix del inet antes del merge y listo.

🌙 *Nyx*

ArtemioPadilla:
## Re-review — all previous issues addressed ✅

Checked the latest force-push (`3624a14`). Everything from the first round is fixed:

**#1 `role-grant` grant history** ✅ — Fresh grants use `INSERT`, extensions use `UPDATE SET revoked_at = expiresAt` only, leaving `granted_at`/`granted_by` untouched.

**#2 `adminClient` naming collision** ✅ — Renamed to `serviceClient` in `auth.ts`.

**#3 `user_roles` RLS policies** ✅ — `user_roles_admin_or_self_read` and `user_roles_no_self_write` added. `admin_audit` gets the same pattern.

**#4 Stub route keys** ✅ — All 25 route keys in `utils.ts` with full EN/ES slugs.

**#5 `x-forwarded-for` → `inet`** ✅ — Confirmed in `index.ts`: the xFF header is split on comma, first entry taken and trimmed before passing to `insertAuditRow`. Comment in the code explains the proxy case.

**Bonus**: `researcher` role documented in `console-tabs.ts` as an RLS gate role, not a console role.

---

All 5 issues resolved. **LGTM — ready to merge.** 🦅

PR #71, opened 2026-04-29
fix(profile): 403 on profile/settings — drop client-set updated_at + server-side touch trigger

ArtemioPadilla:
## Summary

User-reported prod bug:
```
PATCH /rest/v1/users?id=eq.<self> → 403 Forbidden
```
when saving any change in `/profile/settings/profile/`.

## Root cause

The PATCH payload included `updated_at: new Date().toISOString()`. The column-level GRANT on `public.users` (the same one inventoried in PR #69) intentionally excludes `updated_at` because it's an audit column — clients shouldn't dictate it. PostgREST rejects the **entire** UPDATE when any patched column is outside the GRANT, so even valid edits (`username`, `bio`, `display_name`, etc.) silently couldn't be saved.

Same latent bug existed in two more places that hadn't been hit yet:
- `PrivacyMatrix.astro` — both the `supabase-js update` AND the `keepalive` PATCH that fires on `beforeunload` would have 403'd the next time someone toggled a privacy setting
- `AdminExpertsView.astro` — would fail once it migrates from the service-role admin function to direct PostgREST

## Two-part fix

**1) DB owns the timestamp.** New `tg_users_touch_updated_at` BEFORE UPDATE trigger sets `NEW.updated_at = now()` on every row update. Without this, dropping the client write would silently rot the column.

```sql
CREATE OR REPLACE FUNCTION public.tg_users_touch_updated_at()
RETURNS trigger LANGUAGE plpgsql AS $$
BEGIN NEW.updated_at := now(); RETURN NEW; END;
$$;
DROP TRIGGER IF EXISTS users_touch_updated_at ON public.users;
CREATE TRIGGER users_touch_updated_at
  BEFORE UPDATE ON public.users
  FOR EACH ROW EXECUTE FUNCTION public.tg_users_touch_updated_at();
```

**2) Client stops sending it.** Removed from `ProfileEditForm.astro`, both call sites in `PrivacyMatrix.astro`, and `AdminExpertsView.astro`. Audit columns shouldn't come from the client anyway — defense in depth.

**3) Doc updated.** `docs/specs/infra/users-column-grants.md` now explicitly calls out `tg_users_touch_updated_at` as the writer for `updated_at`, with a one-line "this exact bug surfaced when ProfileEditForm was setting it from the client → 403" so the next person editing the GRANT sees the trap.

## Operator action

Automatic. The SQL change is in `supabase-schema.sql` so `db-apply.yml` fires on the merge push. `CREATE OR REPLACE` + `DROP TRIGGER IF EXISTS … ; CREATE TRIGGER …` are idempotent.

## Test plan

- [x] `npm run typecheck` — 0 errors
- [x] `npm run test` — 454 passing
- [x] `npm run build` — clean
- [ ] Post-merge manual: edit username/bio/display_name in `/profile/settings/profile/` → save → should succeed; row's `updated_at` should advance to now (verifiable via Supabase dashboard or `select updated_at from users where id = …`).

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -3125,6 +3125,27 @@ CREATE TRIGGER obsreact_notify_trigger
   AFTER INSERT ON public.observation_reactions
   FOR EACH ROW EXECUTE FUNCTION public.tg_obsreact_notify();
 
+-- 15) Audit columns. The column-level GRANT on `public.users`
+-- intentionally does NOT include `updated_at` (it's an audit column
+-- — clients shouldn't dictate it). Without this BEFORE UPDATE trigger
+-- the column would silently rot. The trigger sets NEW.updated_at on
+-- every row update so the DB owns the timestamp.
+--
+-- See docs/specs/infra/users-column-grants.md for the column inventory.
+CREATE OR REPLACE FUNCTION public.tg_users_touch_updated_at()
+RETURNS trigger
+LANGUAGE plpgsql AS $$
+BEGIN
+  NEW.updated_at := now();
+  RETURN NEW;
+END;
+$$;
+
+DROP TRIGGER IF EXISTS users_touch_updated_at ON public.users;
+CREATE TRIGGER users_touch_updated_at
+  BEFORE UPDATE ON public.users
+  FOR EACH ROW EXECUTE FUNCTION public.tg_users_touch_updated_at();
+
 -- ═════════════════════════════════════════════════════════════════════
 -- Module 27 — Expertise Legends (regional rankings)
 -- Issue #47 — "Top identificador de Fabaceae en Oaxaca"
--- a/docs/specs/infra/users-column-grants.md
+++ b/docs/specs/infra/users-column-grants.md
@@ -50,7 +50,8 @@ from `authenticated` is denied.
 | Column | Writer | Why locked |
 |---|---|---|
 | `id` | `auth.users` FK | Primary key, immutable |
-| `created_at` / `updated_at` / `joined_at` | trigger / default | Audit columns |
+| `created_at` / `joined_at` | column DEFAULT now() on INSERT | Audit columns; immutable after creation |
+| `updated_at` | `tg_users_touch_updated_at` BEFORE UPDATE trigger | Audit column; trigger sets `NEW.updated_at = now()` so clients shouldn't touch it (and the GRANT prevents it). Bug surfaced when ProfileEditForm was setting it from the client → 403. |
 | `is_expert` | `expert_applications` admin flow + `is_expert_in()` SQL | Self-elevation would let anyone weight their votes |
 | `credentialed_researcher` / `credentialed_at` / `credentialed_by` | admin grant | Unlocks precise coords on sensitive species |
 | `karma_total` / `karma_updated_at` / `vote_count` | karma engine triggers (SECURITY DEFINER) | Self-bump would game leaderboards + vote weighting |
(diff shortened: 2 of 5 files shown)


ArtemioPadilla:
Fix correcto y bien ejecutado.

**Root cause:** el análisis es exacto — PostgREST rechaza el UPDATE completo cuando cualquier columna del payload está fuera del GRANT, no solo esa columna. Es el comportamiento esperado de PostgreSQL column-level security, pero es una trampa no obvia porque el error es 403 en toda la operación, no "columna X denegada".

**El trigger `tg_users_touch_updated_at`:**
`BEFORE UPDATE … SET NEW.updated_at = now(); RETURN NEW;` es el patrón estándar y correcto. No necesita `SECURITY DEFINER` porque solo toca la columna del row que ya está siendo actualizado — no hay cross-table access. El `DROP TRIGGER IF EXISTS … ; CREATE TRIGGER …` hace el bloque idempotente para el CI.

**Defensa en profundidad bien aplicada:** limpiar los tres call sites (`ProfileEditForm`, `PrivacyMatrix` × 2, `AdminExpertsView`) aunque solo el primero estaba roto es la decisión correcta — los otros tres hubieran fallado en cuanto algún path los alcanzara con usuario `authenticated` (en vez de service-role).

**Nota para el backlog:** la Nyx señaló en PR #68 que el patrón de column-level GRANT se vuelve frágil conforme crecen los triggers. El doc update en `users-column-grants.md` (separar `created_at`/`joined_at` de `updated_at` y nombrar el trigger explícitamente) es exactamente el inventario que se pedía. El next trigger que toque `users` debería actualizar esa tabla primero.

✅ Aprobado, listo para merge.

Pamela-ruiz9 (approved):
## Review — PR #71 fix(profile): 403 on profile/settings

### Root cause analysis

This is exactly the right diagnosis. PostgREST rejects the **entire** UPDATE when any column in the payload is outside the GRANT — not just that column. So the client setting `updated_at: new Date().toISOString()` was silently blocking every profile save since the column-level GRANT shipped (PR #69). Two more latent instances correctly caught in `PrivacyMatrix.astro` and `AdminExpertsView.astro` before they were hit in prod.

### Two-part fix is correct

**`tg_users_touch_updated_at` BEFORE trigger:** Right layer. The DB owns audit columns — if the client is not allowed to write `updated_at`, the column still needs to advance on every row change. `BEFORE UPDATE FOR EACH ROW` setting `NEW.updated_at = now()` is the standard pattern. `DROP TRIGGER IF EXISTS … ; CREATE TRIGGER …` makes it idempotent — safe for re-apply.

**Removing from client payloads:** Defense in depth. The trigger alone would have fixed the bug (removing the forbidden column), but keeping `updated_at` in the payload would still surface a 403 on any future GRANT-change that adds the column back. Removing it from the client removes the coupling entirely.

**Doc update in `users-column-grants.md`:** This is the detail that pays forward. Explicitly naming the trigger as the writer, and leaving a one-line "this exact bug surfaced when ProfileEditForm was setting it from the client → 403" means the next person editing the GRANT sees the trap before falling into it.

### One small note (non-blocking)

The comment in `PrivacyMatrix.astro` reads:
```
// updated_at handled by users_touch_updated_at trigger.
```
while `ProfileEditForm.astro` has a fuller comment:
```
// updated_at is bumped server-side by the users_touch_updated_at trigger; clients can't write to it (column-level GRANT).
```
Not a bug, but consistency on the comment style across the three files would make a grep for "updated_at" more uniform. Low priority.

### On `db-apply.yml` auto-fire

`CREATE OR REPLACE FUNCTION` + `DROP TRIGGER IF EXISTS … CREATE TRIGGER` in `supabase-schema.sql` means the schema apply is idempotent and the trigger lands automatically on merge. No manual step needed — this is exactly the value of the hardened apply from PR #61.

---

**+28/-5 in 5 files, 454 passing. Clean fix. 🌙**

PR #77, opened 2026-04-29
feat(console): PR5 — Moderator console (flag queue + comments + soft-bans + role-aware overview)

ArtemioPadilla:
## Summary

The moderator role goes from "all stubs" to "all functional except disputes." Brings the console fully aligned with the architectural rule:

> "deberíamos de tener todo en la UI si eres admin, si no eres admin ni moderador nada. si eres moderador solo lo necesario."

Admin sees their 15 tabs (10 functional after this), moderator sees their 5 (4 functional + 1 stub), no-role sees no console pill.

### What ships

**Moderator overview** (role-aware, shares `/console/` route)
- The `/console/` page now dispatches to ConsoleOverviewView, ConsoleModOverviewView, or (later) ConsoleExpertOverviewView based on the active role pill.
- Mod overview: open reports counter, comments hidden 24h, active bans, mean response-time on resolved reports last 7d.

**Flag queue** (`/console/flag-queue/`)
- Master-detail of `public.reports` rows — community-submitted flags from m26's `report` Edge Function
- Filter by status / target_type / reason; default = open + triaged
- Action chips: Triage, Resolve, Dismiss — slide-overs with required reason

**Comments** (`/console/comments/`)
- Master-detail of `public.observation_comments` — m26 thread schema
- Filters: deleted state, locked state, observation_id, author, body search
- Action chips: Hide / Unhide / Lock / Unlock

**Soft-bans** (`/console/bans/`)
- New `public.user_bans` table — temporary suspensions with `expires_at` (nullable = indefinite) and soft-revoke (`revoked_at`)
- Master-detail of active bans (toggleable to all)
- "Ban a user" affordance with duration select (1h / 24h / 7d / 30d / indefinite)
- Action chip: Unban (lifts active bans early, sets `revoked_at`/`revoked_by`/`revoke_reason`)

### Schema additions (idempotent)

- `public.user_bans` table + active-row index + RLS (`user_bans_mod_read` + `user_bans_no_client_write`)
- `public.is_user_banned(uuid)` SECURITY DEFINER function for downstream RLS predicates that want to lock out banned users from acting
- `reports_mod_read` + `comments_mod_read` policies (both include moderator OR admin in USING — admins see full queues without holding moderator)
- `observation_comments.locked boolean DEFAULT false`
- 4 new audit_op values: `report_triaged`, `report_resolved`, `report_dismissed`, `comment_unhide`
- Sentinel updated with `user_bans` + `is_user_banned`

[...]

PR #79, opened 2026-04-29
fix(console): PR #77 review — unban consistency + double-ban guard + per-command policies + locked sentinel

ArtemioPadilla:
## Summary

Four review-comment fixes from Pamela-ruiz9 (Nyx, APPROVED) + ArtemioPadilla (APPROVED) on PR #77 (Moderator console, already merged).

### 1. `user-unban.ts` — consistency check (Nyx, **real correctness bug**)

The handler updated by `ban_id` only. A moderator could pass mismatched `ban_id` + `target_user_id`, the UPDATE would still succeed, and the audit row's `target.id` would point at the wrong user — the audit chain would lie.

Both the BEFORE fetch and the UPDATE now carry `.eq('id', payload.ban_id).eq('user_id', payload.target_user_id)`. If either fails to match, the handler throws `'unban: ban_id does not belong to target_user_id'` and no audit row gets written. Audit chain integrity preserved.

### 2. `user-ban.ts` — idempotent re-ban (yours)

Double-click on the ban button (or two moderators acting simultaneously) inserted two active rows for the same user. `is_user_banned()` still functioned, but the bans tab showed duplicates and the audit log had two creation events for the same effective state.

Now checks for an active ban (not revoked, not expired) before inserting. If one exists, returns it as `before` + `after` and short-circuits — no duplicate row. The audit row is still written for both attempts so monitoring sees the activity. Future v1.x might extend `expires_at` if the new request has a longer duration; for v1 we no-op.

### 3. `user_bans_no_client_write` — split per-command (yours, clarity)

The single `FOR ALL USING(false) WITH CHECK(false)` was functionally correct (writes from `authenticated` blocked, SELECT delegated to `user_bans_mod_read` via the OR semantics) but the name applied to a `FOR ALL` policy was misleading.

Replaced with three explicit per-command policies: `user_bans_no_client_insert`, `user_bans_no_client_update`, `user_bans_no_client_delete`. The legacy name is DROP'd first for replay-safety against the already-deployed prod DB.

### 4. Sentinel — `observation_comments.locked` column check (yours, defensive)

`ALTER TABLE … ADD COLUMN IF NOT EXISTS locked` is safe in Postgres so risk is low, but a partial migration could land the table without the column and leave the moderator UI silently degraded. Sentinel now does an `information_schema.columns` lookup using the established `array_append(missing, …)` pattern.

## Deferred (correctly per reviewer notes)

- pgTAP RLS policy assertions — already on PR8 list (CORS + rate limit + pgTAP)
- `locked` enforcement on insert (block new comments on locked threads) — v1.1 follow-up

## Test plan

- [x] `npm run typecheck` — zero errors
- [x] `npm run test` — 459/459 passed
- [x] `npm run build` — 122 pages, zero errors
- [x] `npm run test:e2e -- console-smoke.spec.ts` — 28/28 passed
- [ ] **Post-merge db-apply** — sentinel will verify `locked` column is present on prod (it should be, since PR #77's ALTER TABLE landed; this is a regression guard going forward)

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -3473,10 +3473,15 @@ CREATE POLICY user_bans_mod_read ON public.user_bans
   );
 
 -- No client-side writes — service role only via Edge Function.
+-- Three separate per-command policies replace the previous FOR ALL form for
+-- clarity; Postgres processes them independently per statement type.
 DROP POLICY IF EXISTS user_bans_no_client_write ON public.user_bans;
-CREATE POLICY user_bans_no_client_write ON public.user_bans
-  FOR ALL TO authenticated
-  USING (false) WITH CHECK (false);
+DROP POLICY IF EXISTS user_bans_no_client_insert ON public.user_bans;
+DROP POLICY IF EXISTS user_bans_no_client_update ON public.user_bans;
+DROP POLICY IF EXISTS user_bans_no_client_delete ON public.user_bans;
+CREATE POLICY user_bans_no_client_insert ON public.user_bans FOR INSERT TO authenticated WITH CHECK (false);
+CREATE POLICY user_bans_no_client_update ON public.user_bans FOR UPDATE TO authenticated USING (false);
+CREATE POLICY user_bans_no_client_delete ON public.user_bans FOR DELETE TO authenticated USING (false);
 
 GRANT SELECT ON public.user_bans TO authenticated;
 GRANT SELECT, INSERT, UPDATE, DELETE ON public.user_bans TO service_role;
--- a/scripts/db-sentinel.sql
+++ b/scripts/db-sentinel.sql
@@ -51,6 +51,16 @@ BEGIN
   -- Console PR5 — moderator surface
   IF NOT EXISTS (SELECT 1 FROM pg_tables WHERE schemaname = 'public' AND tablename = 'user_bans')              THEN missing := array_append(missing, 'public.user_bans'); END IF;
 
+  -- Module 26 — observation_comments.locked column added in PR #77
+  IF NOT EXISTS (
+    SELECT 1 FROM information_schema.columns
+    WHERE table_schema = 'public'
+      AND table_name = 'observation_comments'
+      AND column_name = 'locked'
+  ) THEN
+    missing := array_append(missing, 'public.observation_comments.locked');
+  END IF;
+
   -- Functions used by RLS predicates and Edge Functions
   IF NOT EXISTS (SELECT 1 FROM pg_proc WHERE proname = 'has_role')             THEN missing := array_append(missing, 'public.has_role()'); END IF;
   IF NOT EXISTS (SELECT 1 FROM pg_proc WHERE proname = 'recompute_consensus')  THEN missing := array_append(missing, 'public.recompute_consensus()'); END IF;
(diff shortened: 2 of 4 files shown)


ArtemioPadilla:
Los 4 review comments del PR #77 atendidos correctamente. ✅

**`user-unban.ts` — consistency check (Nyx):**
Doble `.eq('id', ban_id).eq('user_id', target_user_id)` tanto en el SELECT como en el UPDATE, más el check de `!updated || updated.length === 0` post-UPDATE. La cadena de auditoría no puede mentir: si `ban_id` no pertenece a `target_user_id`, el handler lanza antes de que se escriba el audit row.

**`user-ban.ts` — dedup:**
La query `.is('revoked_at', null).or('expires_at.is.null,expires_at.gt.<now>')` es el filtro correcto para "ban activo". El short-circuit con `before = after = existing` aún dispara el audit row del intento — correcto para trazabilidad. El comentario sobre v1.x (extend `expires_at` si la nueva duración es mayor) documenta el tradeoff sin implementarlo prematuramente.

**Políticas per-comando:**
`DROP POLICY IF EXISTS user_bans_no_client_write` primero — replay-safe contra la prod donde ya existe. Los tres nombres nuevos (`_insert`, `_update`, `_delete`) son explícitos. Nota: `FOR INSERT` usa solo `WITH CHECK`, `FOR UPDATE` y `FOR DELETE` usan `USING` — sintaxis correcta por comando en Postgres.

**Sentinel `observation_comments.locked`:**
`information_schema.columns` con `table_schema + table_name + column_name` es la consulta correcta para verificar existencia de columna. Consistente con el patrón `array_append(missing, …)` del resto del sentinel.

✅ Listo para merge.

Pamela-ruiz9 (approved):
## Review — PR #79 fix(console): PR #77 review follow-ups

All four items from the PR #77 review addressed correctly.

### `user-unban.ts` — consistency check (my note, #77)

The fix is correct and goes one step further than the suggestion: both the BEFORE fetch AND the UPDATE now carry the double `.eq()` guard. The extra `if (!updated || updated.length === 0)` after the UPDATE means a race condition (another moderator revokes the ban between the fetch and the UPDATE) also throws instead of silently succeeding. The error message `'unban: ban_id does not belong to target_user_id'` covers both cases — slightly misleading on the race path, but that's a very minor point not worth blocking on.

### `user-ban.ts` — idempotent re-ban

`.maybeSingle()` is the right PostgREST method here (vs `.single()` which would throw on zero rows). The `or('expires_at.is.null,expires_at.gt.' + now)` condition mirrors the `is_user_banned()` predicate — the two stay in sync.

Note: the existing-ban check uses a JS-computed timestamp (`new Date().toISOString()`) rather than a server-side `now()`. In practice the skew is negligible, but if the Edge Function clock drifts it could return a ban as active when the DB's `now()` would consider it expired. Low risk, just noting it.

### RLS policies — per-command split

`DROP POLICY IF EXISTS user_bans_no_client_write` before creating the three new ones handles replay-safety against prod. The per-command forms are clearer and the Postgres behavior is unchanged — three `false` predicates, one per statement type.

### Sentinel — `observation_comments.locked` column

`information_schema.columns` lookup with the established `array_append(missing, …)` pattern is consistent with the rest of the sentinel. This is the first column-level check in the sentinel (vs table/function checks) — good precedent to extend for other critical columns.

---

**+58/-10 in 4 files, 459/459 vitest, 28/28 e2e. All PR #77 review items closed. 🌙**

PR #80, opened 2026-04-29
fix(postgrest): disambiguate users embed on observations (2 surfaces)

ArtemioPadilla:
User-reported error in prod:
```
Could not embed because more than one relationship was found for 'observations' and 'users'
```

**Root cause:** `observations` has two FKs to `users` — the original `observer_id` (NOT NULL, the observer) and `hidden_by` (added by the admin console moderation work for hide audit). PostgREST can't auto-pick when an embed says bare `users(...)`.

**Fix:** alias-style hint `users:observer_id(...)` in:
- `ConsoleObservationsView.astro:219` (admin moderation list)
- `ExploreRecentView.astro:243` (public recent feed)

The alias both names the relationship explicitly and keeps the embed key as `users` so consumer code that reads `row.users` doesn't need any refactoring.

Audited every other `from('observations')` site — none of them embed `users`, so these two were the only affected surfaces.

Verifications: typecheck 0, 454 tests, build clean.

### DIFF
--- a/src/components/ConsoleObservationsView.astro
+++ b/src/components/ConsoleObservationsView.astro
@@ -216,7 +216,7 @@ const tr = t(lang);
 
     let query = supabase
       .from('observations')
-      .select('id, observer_id, observed_at, primary_taxon_id, obscure_level, location_obscured, observer_license, hidden, hidden_reason, hidden_at, hidden_by, sync_status, created_at, taxa(scientific_name), users(username, display_name)')
+      .select('id, observer_id, observed_at, primary_taxon_id, obscure_level, location_obscured, observer_license, hidden, hidden_reason, hidden_at, hidden_by, sync_status, created_at, taxa(scientific_name), users:observer_id(username, display_name)')
       .order('created_at', { ascending: false })
       .limit(50);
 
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
Fix correcto. El alias  en PostgREST es la sintaxis exacta para disambiguar cuando hay múltiples FKs hacia la misma tabla — nombra la relación por la columna FK, y el alias `users` preserva la key en el objeto de respuesta, así `row.users` sigue funcionando sin tocar el código consumidor.

La auditoría de todos los sitios con `from('observations')` vale mencionarla explícitamente — es lo que confirma que no quedan otros rotos.

✅ Aprobado, listo para merge.

Pamela-ruiz9 (approved):
## Review — PR #80 fix(postgrest): disambiguate users embed on observations

Correct diagnosis and minimal fix. `observations` has two FKs to `users` (`observer_id` + `hidden_by` from the admin moderation work) — PostgREST's implicit embed resolver fails ambiguously when both could satisfy `users(…)`. The `users:observer_id(…)` alias syntax is the right PostgREST idiom: it names the FK explicitly, the alias keeps the returned key as `users` so consumer code is unchanged.

Both affected surfaces fixed: the admin moderation list and the public recent feed. The audit of all other `from('observations')` sites confirming none embed `users` is the right completeness check.

One note for awareness (not blocking): `hidden_by` is a FK to `users` and will have the same ambiguity if any future query embeds `users:hidden_by(…)` alongside `users:observer_id(…)` — but that would need an explicit alias too, and the explicit form is already established here. No action needed.

**+2/-2 in 2 files. Clean. 🌙**

PR #272, opened 2026-04-30
fix(rls): add explicit WITH CHECK to obs_owner — resolves 42P17 on observation INSERT

ArtemioPadilla:
## Problem

Observations submitted by users (including field test by Marcela Padilla, 2026-04-30) fail with:

```
POST /rest/v1/observations?on_conflict=id → 500
proxy_status: PostgREST; error=42P17
```

**42P17 = "infinite recursion detected in policy for relation"**

The outbox sync shows a spinner but never resolves. The observation stays in local IndexedDB forever.

## Root cause

`obs_owner` policy was defined as `FOR ALL` with only a `USING` clause and no `WITH CHECK`:

```sql
CREATE POLICY "obs_owner" ON public.observations FOR ALL
  USING ((SELECT auth.uid()) = observer_id);
```

When Postgres evaluates an `INSERT`, permissive policies without explicit `WITH CHECK` fall back to using their `USING` clause for both read and write checks. With 5+ permissive policies on `observations` (`obs_owner`, `obs_public_read`, `obs_credentialed_read`, `obs_collaborator_read`, `obs_admin_full_read`) — some added recently (M28 console, M29 community, M32 multi-provider) — the policy evaluation chain can recursively re-enter itself, triggering 42P17.

## Fix

Add explicit `WITH CHECK` to `obs_owner`:

```sql
CREATE POLICY "obs_owner" ON public.observations FOR ALL
  USING    ((SELECT auth.uid()) = observer_id)
  WITH CHECK ((SELECT auth.uid()) = observer_id);
```

This short-circuits the fallback behavior. INSERT/UPDATE now uses the explicit `WITH CHECK` without needing to re-evaluate the full policy chain.

## Impact

- All observation uploads were failing for authenticated users since the policy accumulation threshold was crossed
- Outbox sync appears to retry infinitely with no user-facing error (just a spinner)

## To apply in prod

After merging, run the db-apply workflow or execute manually in Supabase SQL editor:

```sql
DROP POLICY IF EXISTS "obs_owner" ON public.observations;
CREATE POLICY "obs_owner" ON public.observations FOR ALL
  USING    ((SELECT auth.uid()) = observer_id)
  WITH CHECK ((SELECT auth.uid()) = observer_id);
```

## Verification

```sql
SELECT polname, polcmd, 
       pg_get_expr(polqual, polrelid) AS using_clause,
       pg_get_expr(polwithcheck, polrelid) AS with_check_clause
FROM pg_policy
JOIN pg_class ON pg_class.oid = pg_policy.polrelid
WHERE pg_class.relname = 'observations'
ORDER BY polname;
```

After fix: `obs_owner` should show both `using_clause` AND `with_check_clause` populated.

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -399,7 +399,11 @@ CREATE POLICY "taxa_public_read" ON public.taxa FOR SELECT USING (true);
 -- policy can stay single-table and inexpensive.
 DROP POLICY IF EXISTS "obs_owner" ON public.observations;
 CREATE POLICY "obs_owner" ON public.observations FOR ALL
-  USING ((SELECT auth.uid()) = observer_id);
+  USING    ((SELECT auth.uid()) = observer_id)
+  -- Explicit WITH CHECK prevents Postgres from falling back to the USING clause
+  -- on INSERT/UPDATE, which in complex policy environments can trigger 42P17
+  -- infinite recursion when other permissive policies subquery this table.
+  WITH CHECK ((SELECT auth.uid()) = observer_id);
 
 DROP POLICY IF EXISTS "obs_public_read" ON public.observations;
 CREATE POLICY "obs_public_read" ON public.observations FOR SELECT


Pamela-ruiz9 (commented):
## Review — PR #272 (fix RLS 42P17)

### ✅ The fix is correct

Adding an explicit `WITH CHECK` to the `obs_owner` policy is the right approach. The root cause is well-understood:

When a Postgres RLS policy uses `FOR ALL` with only a `USING` clause, Postgres automatically uses the `USING` expression for both row filtering (SELECT/UPDATE/DELETE) and row validation (INSERT/UPDATE). In a multi-policy environment where other permissive policies subquery the same table, this can cause the planner to evaluate `obs_owner` recursively — triggering `42P17`.

Making the `WITH CHECK` explicit breaks the recursion because the planner now has a distinct expression for each path and doesn't need to expand `USING` into `WITH CHECK` context.

**The `(SELECT auth.uid()) = observer_id` pattern is also correct** — wrapping `auth.uid()` in a subquery prevents it from being re-evaluated per-row in set-returning operations (a known Supabase RLS performance pattern).

---

### ⚠️ One concern: root cause not fully confirmed

The PR description says "other permissive policies subquery this table" is the cause, but doesn't identify which policy triggers the chain. The fix is safe and almost certainly correct, but there's a small risk that:

1. The 42P17 has a different root cause (a recently-added trigger or function calling `observations` indirectly)
2. The fix masks the symptom without removing the underlying recursive policy

**Recommendation:** After applying the fix in prod, run this query in Supabase SQL Editor to confirm no other policies still subquery `observations`:

```sql
SELECT polname, polcmd, pg_get_expr(polqual, polrelid) AS using_expr, pg_get_expr(polwithcheck, polrelid) AS with_check_expr
FROM pg_policy
JOIN pg_class ON pg_class.oid = pg_policy.polrelid
WHERE pg_class.relname = 'observations';
```

If any other policy's `using_expr` references `observations` (or a function that does), that should also be fixed to avoid future regressions.

---

### Minor

- The schema comment is clear and helpful — good practice for the next person who touches this policy
- No test coverage for RLS policies in the repo currently — this is known tech debt, not introduced by this PR

---

### Summary

**LGTM** — safe to apply. The fix is minimal, correct, and doesn't affect any read paths. Applying the SQL directly in Supabase SQL Editor as ArtemIO suggested will unblock Marcela's pending observations immediately.

PR #275, opened 2026-04-30
fix(console): observer_license from users join + entity-browser range guard (#275)

ArtemioPadilla:
### Bug 1 — `/console/observations/` error: `column observations.observer_license does not exist`

**Root cause:** `ConsoleObservationsView` was selecting `observer_license` directly from the `observations` table. The column lives on `users` (set at profile creation per module 08).

**Fix:** Move `observer_license` into the `users:observer_id(...)` embed in the SELECT clause. Update `ObsRow` type and all read sites to use `o.users?.observer_license ?? 'CC BY 4.0'`.

### Bug 2 — `/console/notifications/` entity-browser warning: `a filter.apply returned a non-query value. Active filters: []`

**Root cause:** `entity-browser.ts refresh()` had two branches — `filtersChanged` used `select(..., { count: 'exact' })` while the else branch omitted it. After the `as unknown as SupabaseQuery` cast, the two resulting builder shapes differed internally. When the notifications tab loads for the first time with no active filters, `filtersChanged = true` on first call, producing a builder the `typeof q.range !== 'function'` guard detected as broken.

**Fix:** Unify both branches to always use `{ count: 'exact' }`. The count is already gated behind `if (filtersChanged)" at use-site (~line 478), so behavior is unchanged. The spurious warning disappears.

### DIFF
--- a/src/components/ConsoleObservationsView.astro
+++ b/src/components/ConsoleObservationsView.astro
@@ -174,7 +174,7 @@ const tr = t(lang);
     primary_taxon_id: string | null;
     obscure_level: string;
     location_obscured: unknown;
-    observer_license: string;
+    observer_license: string | null;
     hidden: boolean;
     hidden_reason: string | null;
     hidden_at: string | null;
@@ -182,7 +182,7 @@ const tr = t(lang);
     sync_status: string;
     created_at: string;
     taxa: { scientific_name: string } | null;
-    users: { username: string | null; display_name: string | null } | null;
+    users: { username: string | null; display_name: string | null; observer_license: string | null } | null;
   };
 
   function show(id: string) { document.getElementById(id)?.classList.remove('hidden'); }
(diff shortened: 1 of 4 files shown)


Pamela-ruiz9 (commented):
## Review — PR #275 (fix console bugs)

### Bug 1: `observer_license` join fix ✅

The root cause is correctly identified and the fix is right: `observer_license` belongs to `users`, not `observations`. The changes are:

1. **Type update** — `observer_license` moved from `observations` row type to `users` nested object. `string | null` is more accurate than `string` (not every user will have set a license). ✅
2. **SELECT clause** — removed from observations columns, added to `users:observer_id(...)` join. ✅
3. **License filter** — changed from `.eq('observer_license', license)` to `.eq('users.observer_license', license)`. This should work with PostgREST embedded resource filtering, but worth noting: PostgREST v11+ supports `embedded_filter` but the syntax may need to be `.eq('users.observer_license', value)` or require a different approach. If the filter doesn't work in prod, the fallback is client-side filtering. Non-blocking for this PR. ✅
4. **Rendering** — `o.observer_license` → `o.users?.observer_license ?? 'CC BY 4.0'`. The fallback default of `'CC BY 4.0'` is reasonable and consistent with what Rastrum sets at user creation. ✅

**One minor concern:** the license action modal reads `allObs.find(x => x.id === selectedId)?.users?.observer_license ?? 'CC BY 4.0'` — if the update action then calls a PATCH on `observations.observer_license`, that will fail too (same column-doesn't-exist error). I don't see the PATCH handler in this diff so assuming it already updates `users.observer_license`. Worth verifying.

---

### Bug 2: entity-browser range guard false positive ✅

The fix is clean: unify both branches of the query builder into one, always passing `{ count: 'exact' }`. This ensures `.range()` is always available on `q` regardless of the `filtersChanged` flag.

**Why this works:** In Supabase JS v2, `select(clause, { count: 'exact' })` returns a `PostgrestTransformBuilder` with `.range()` available. The conditional branching was creating two different builder shapes; unifying them fixes the type mismatch.

The comment explaining *why* the two branches caused the issue is helpful — good docs hygiene.

**Minor:** `count: 'exact'` is now always requested even when filters haven't changed. For large tables this adds a small overhead (COUNT query runs alongside every page load, not just on filter change). For a console/admin view this is acceptable — admins are few and the count is useful for pagination anyway. Non-blocking. ✅

---

### Summary

**LGTM** — both fixes are correct, minimal, and well-explained. Safe to merge. The only open question is whether the license PATCH action in the console also needs updating — if it patches `observations.observer_license` directly, that will still 500. Can be verified quickly in the admin console after merge.

PR #276, opened 2026-04-30
chore: tech debt cleanup — isEs2, classic sunset, AudioPlayer dedup

ArtemioPadilla:
### 1. `isEs2` residual removed (ObservationForm.astro)
- Removed the duplicate `const isEs2 = ...` declaration (line ~1575) that shadowed the already-in-scope `isEs` from line 717
- Replaced all 5 `isEs2` references with `isEs`

### 2. Sunset banner for classic form pages
- Added deprecation banner to `/en/observe/classic` and `/es/observar/clasico`
- Banner text: ⚠️ Classic form will be removed on 2026-06-30. Use the new /observe instead.
- Added `<!-- SUNSET: 2026-06-30 — remove this page and redirect to /observe -->` HTML comment at top of both files

### 3. AudioPlayer HTML deduplication (share/obs/index.astro)
- Extracted the ~90-line inline template literal into a `buildAudioPlayerHTML(audioUrl, mimeType, obsIdAttr, lang)` helper function
- The `.map()` call now delegates to this function — clean, DRY, and easier to maintain
- Added `// TODO: when Astro supports SSR hydration in this context, use AudioPlayer.astro directly`

### 4. `observer_license` reference audit ✅
- Confirmed all reads of `observer_license` are from the `users` table (via JOIN), not `observations`
- No changes needed

---

**CI:** typecheck ✅ (0 errors) · vitest ✅ (701/701 passed)

### DIFF
--- a/src/components/ObservationForm.astro
+++ b/src/components/ObservationForm.astro
@@ -1572,7 +1572,7 @@ const idAgainFailedLabel   = observeStrings.id_again_failed         ?? (isEs ? '
   let durationCutoffTimer: number | null = null;
 
   const AUDIO_MAX_SECS = 30;
-  const isEs2 = (typeof document !== 'undefined') && document.documentElement.lang === 'es';
+  
 
   function repaintAudioGrid() {
     if (!audioGrid) return;
@@ -1607,7 +1607,7 @@ const idAgainFailedLabel   = observeStrings.id_again_failed         ?? (isEs ? '
   if (!audioSupported()) {
     if (audioRecBtn) audioRecBtn.disabled = true;
     if (audioErrEl) {
-      audioErrEl.textContent = isEs2 ? 'La grabación de audio no está disponible en este navegador.' : "Audio recording isn't supported on this browser.";
+      audioErrEl.textContent = isEs ? 'La grabación de audio no está disponible en este navegador.' : "Audio recording isn't supported on this browser.";
       audioErrEl.classList.remove('hidden');
     }
   }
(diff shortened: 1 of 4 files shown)


Pamela-ruiz9 (commented):
## Review — PR #276 (tech debt cleanup)

### ✅ What's well done

**`isEs2` removal** — Clean and correct. All 5 references replaced with the `isEs` that already exists in the outer Astro frontmatter scope. The empty line left where the `const isEs2` was is harmless. No behavioral change, just deduplication. ✅

**Sunset banners** — Good approach:
- Amber warning banner visible to users on both classic pages ✅
- `<!-- SUNSET: 2026-06-30 -->` comment above the frontmatter for grepping in the future ✅
- June 30 gives users 2 months — reasonable ✅

**Minor:** The Spanish sunset banner still says "Classic form will be removed" in English. Should probably say "El formulario clásico se eliminará el 2026-06-30. Usa el nuevo /observar."

**AudioPlayer extraction to `buildAudioPlayerHTML()`** — This is the right direction:
- Function signature is clear: `(audioUrl, mimeType, obsIdAttr, lang) → string` ✅
- The TODO comment for SSR migration is helpful — makes it obvious this is an intermediate step ✅
- Template string stays the same length; no logic was lost ✅

**One concern:** `buildAudioPlayerHTML` is defined inside the `<script>` tag (Astro client-side script), but the function also references the large HTML block with dynamic IDs. This is fine as-is, but if this function ever needs to be shared (e.g., used by a different page), it will need to be extracted to a `.ts` file. The TODO acknowledges this. Not blocking.

---

### ⚠️ Minor issues

1. **Spanish sunset banner text is in English** (line ~12 of `clasico.astro`):
   ```
   ⚠️ Classic form will be removed on 2026-06-30.
   ```
   Should be:
   ```
   ⚠️ El formulario clásico se eliminará el 2026-06-30. Usa el nuevo /observar.
   ```

2. **`isEs2` removal leaves a blank line** — cosmetic, harmless. Could be removed in a followup.

3. **`buildAudioPlayerHTML` uses string interpolation for `audioUrl`** without sanitization. If `af.url` ever contains a `"` character (e.g., a URL with an encoded quote), the `data-audio-url="${audioUrl}"` attribute injection could break. In practice R2 URLs don't contain quotes, but worth noting for security hygiene. The original code had the same issue so this isn't a regression.

---

### Summary

**LGTM with the minor Spanish text fix** — the debt is real, the fixes are correct, and the sunset dates are reasonable. The AudioPlayer extraction is the right intermediate step. Fix the Spanish text and merge.

PR #321, opened 2026-05-01
fix(rls): tighten project_members_read to prevent 42P17 recursion

ArtemioPadilla:
## Summary

Tightens the `project_members_read` RLS policy to remove the `visibility = 'public'` subquery path that caused a `projects ↔ project_members` recursive policy evaluation (42P17).

## Change

`docs/specs/infra/supabase-schema.sql` — `project_members_read` policy:

**Before:** members could see rows if the project is public OR they are the owner  
**After:** members can see rows only if they are the owner (or their own membership row via the `user_id = auth.uid()` arm)

Project owners still see all members via `project_members_owner_write`. Public project membership visibility is a v1.1 follow-up once a non-recursive read path is designed.

## Why

The cycle was:  
`UPDATE projects` → `projects_public_read` → `project_members_read` → `SELECT FROM projects` → loop.

Related: #312 (SECURITY DEFINER helpers), #320 (observation-chain EXISTS fixes).

## Tested

- Idempotent SQL — safe to replay with `make db-apply`
- Note: also contains `.github/workflows/codeql.yml` which is already in main (no-op)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,48 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'  # Weekly on Monday
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      packages: read
+      actions: read
+      contents: read
+
+    strategy:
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
Superseded by #301 (`is_project_owner()` SECURITY DEFINER helper) and #320 (root-cause elimination of all `IN (SELECT FROM observations)` recursion patterns). The fix here would re-introduce the inline subquery into `projects` that #320 intentionally removed. Closing as redundant.

PR #330, opened 2026-05-01
fix(birdnet+audio): sigmoid confidence, persist client ID, audio MIME, M32 source enum

### LINKED ISSUE #2: bug(mobile): header navigation renders twice on mobile

ArtemioPadilla:
## Summary
Fixes the BirdNET / audio-observation bugs the user surfaced:

### 1. Confidence card showed >100% (e.g. \"189%\")
BirdNET-Lite v2.4 emits raw logits, not probabilities. Apply \`sigmoid()\` before returning so the value lands in \`[0,1]\`. Also satisfies the \`identifications.confidence CHECK BETWEEN 0 AND 1\` schema constraint that would have rejected fix #2's insert.

### 2. Saved observations rendered as \"Unknown species\"
\`syncOne()\` was discarding the client's accepted identification and queueing the server-side cascade — but the cascade has no audio identifier (BirdNET is client-only), so \`public.identifications\` stayed empty forever. Now: when a client identification is present and accepted, persist it directly to \`public.identifications\` (\`is_primary: true\`) and skip the cascade. Server cascade still runs as backstop on persist failure.

### 3. Share page showed broken-image card instead of audio player
When \`MediaRecorder\` produces a Blob with empty \`type\` (notably iOS Safari), the save code's \`pf.file.type || 'application/octet-stream'\` yielded the generic MIME, and \`sync.ts\`'s prefix chain defaulted \`media_files.media_type\` to \`'photo'\`. The share page then rendered the audio URL as \`\` (broken icon) and the audio player div stayed hidden. Backfill MIME from \`pf.kind\` when \`pf.file.type\` is empty.

### 4. Schema source enum was missing 9 sources
Extends \`identifications.source\` CHECK from the original 5 server-cascade values to cover all real producers:
- **Server cascade** (kept): \`plantnet\`, \`claude_haiku\`, \`claude_sonnet\`, \`onnx_offline\`, \`human\`
- **Client-side identifiers**: \`birdnet_lite\`, \`onnx_efficientnet_lite0\`, \`camera_trap_megadetector\`, \`phi_vision\`
- **M32 multi-provider vision**: \`bedrock\`, \`openai\`, \`azure_openai\`, \`gemini\`, \`vertex_ai\` (matches the literals hardcoded in \`vision-provider.ts\`)

Idempotent \`DROP CONSTRAINT IF EXISTS\` + \`ADD CONSTRAINT\` migration appended to \`schema.sql\`.

## Known follow-ups (not in this PR)
- Existing rows mis-classified as \`media_type='photo'\` need manual fix or re-upload — code change only prevents new occurrences
- \`AnthropicProvider\` in \`vision-provider.ts\` hardcodes \`'claude_haiku'\` as source regardless of model; should switch on \`this.cred.model\`. Worth a follow-up issue
- Per-media-file BirdNET segments (currently only one identification per observation)

## Test plan
- [x] All 730 unit tests pass + typecheck clean
- [ ] \`make db-apply\` succeeds against an existing DB (idempotent DROP+ADD)
- [ ] After deploy: record bird call → BirdNET shows confidence ≤ 100%
- [ ] After deploy: save audio observation → \`/share/obs/?id=…\` shows audio player + spectrogram, identified species (not \"Unknown\"), reachable BirdNET segments overlay
- [ ] M32 multi-provider identifications (Bedrock / OpenAI / Vertex etc.) successfully persist instead of failing the CHECK constraint

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -291,7 +291,10 @@ CREATE TABLE IF NOT EXISTS public.identifications (
   scientific_name text,                    -- denormalized, stored at ID time
   confidence      numeric CHECK (confidence BETWEEN 0 AND 1),
   source          text NOT NULL
-                  CHECK (source IN ('plantnet','claude_haiku','claude_sonnet','onnx_offline','human')),
+                  CHECK (source IN (
+                    'plantnet','claude_haiku','claude_sonnet','onnx_offline','human',
+                    'birdnet_lite','onnx_efficientnet_lite0','camera_trap_megadetector','phi_vision'
+                  )),
   raw_response    jsonb,                   -- full API response
   is_primary      boolean NOT NULL DEFAULT true,
   is_research_grade boolean DEFAULT false,
@@ -6636,3 +6639,19 @@ END;
 $$;
 
 GRANT EXECUTE ON FUNCTION public.merge_user_accounts(uuid, uuid, uuid, text) TO service_role;
+
+-- ═══════════════════════════════════════════════════════════════════════════
+-- Migration: extend identifications.source CHECK to cover client-side
+-- identifiers (BirdNET-Lite audio, EfficientNet-Lite0 images, MegaDetector
+-- camera-trap, Phi-3.5-vision). Without this, the sync layer's direct
+-- insert of a client identification (added 2026-05-01 to fix audio
+-- observations landing as "Unknown species") fails the CHECK constraint.
+-- ═══════════════════════════════════════════════════════════════════════════
+ALTER TABLE public.identifications
+  DROP CONSTRAINT IF EXISTS identifications_source_check;
+ALTER TABLE public.identifications
+  ADD CONSTRAINT identifications_source_check
+  CHECK (source IN (
+    'plantnet','claude_haiku','claude_sonnet','onnx_offline','human',
+    'birdnet_lite','onnx_efficientnet_lite0','camera_trap_megadetector','phi_vision'
+  ));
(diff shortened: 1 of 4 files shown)


Pamela-ruiz9 (approved):
LGTM ✅

Cuatro bugs reales de producción resueltos correctamente:

1. **Sigmoid fix** — logits crudos de BirdNET v2.4 → [0,1] antes de guardar. Necesario también para satisfacer el CHECK de la DB. ✅
2. **BirdNET identifications persistidas** — el cascade server-side no tiene BirdNET, por eso el nombre nunca aparecía. Persiste directo con is_primary:true cuando hay identificación cliente aceptada, cascade como fallback. Correcto. ✅
3. **MIME iOS Safari** — MediaRecorder con type vacío → fallback desde pf.kind. El bug del audio renderizando como imagen rota era el síntoma directo. ✅
4. **source enum expandido** — 9 fuentes reales que ya existían en el código pero no en el constraint. Migración idempotente. ✅

Follow-ups documentados en el PR body — buena práctica.

CI verde 7/7 ✅

Aprobado desde Nyx 🌙

PR #352, opened 2026-05-01
fix(identifiers): read source from Edge Function response (#332)

### LINKED ISSUE #332: fix(vision-provider): AnthropicProvider hardcodes source as "claude_haiku" regardless of model
## Bug
\`AnthropicProvider.identify()\` in \`supabase/functions/_shared/vision-provider.ts:171\` always tags its \`VisionResult\` as:
\`\`\`ts
return toVisionResult(parseModelJson(text), 'claude_haiku', json);
\`\`\`

…regardless of which model is configured. So if a sponsor (or future personal user — see #_) sets up an Anthropic credential with \`model: 'claude-sonnet-4-5'\` or \`'claude-opus-4'\`, the resulting \`identifications.source\` row will be misleadingly recorded as \`'claude_haiku'\`.

## Impact
- **Audit trail wrong** — admin console / cost dashboards attribute spend to the wrong model
- **Cost-per-ID bookkeeping wrong** — Sonnet ≈ 5× Haiku, Opus ≈ 25× Haiku; analytics will look unrealistic
- **Cascade ordering wrong** — \`runCascade\` sorts by \`cost_per_id_usd\`; if every Anthropic call shows up as Haiku-priced, the ordering decision is corrupted

## Fix
Switch on the model prefix to emit the right source:
\`\`\`ts
function anthropicSourceFromModel(model: string): IDSource {
  if (model.startsWith('claude-opus'))   return 'claude_opus';
  if (model.startsWith('claude-sonnet')) return 'claude_sonnet';
  if (model.startsWith('claude-haiku'))  return 'claude_haiku';
  return 'claude_haiku'; // safe default
}
\`\`\`

Requires extending \`identifications.source\` CHECK enum to include \`'claude_opus'\` (already includes \`claude_haiku\` and \`claude_sonnet\`). Same pattern as PR #330's \`DROP CONSTRAINT IF EXISTS … ADD CONSTRAINT\` migration.

The same pattern likely applies to Bedrock when it routes Anthropic models — \`bedrockModelId()\` translates \`claude-haiku-4-5\` → \`us.anthropic.claude-haiku-4-5-v1:0\`. The \`source: 'bedrock'\` tag throws away the model dimension; consider \`'bedrock_claude_haiku'\` etc., or add a \`model\` column to \`identifications\`.

## Refs
- Surfaced 2026-05-01 during PR #330 review
- File: \`supabase/functions/_shared/vision-provider.ts:171,217\`

ArtemioPadilla:
Fixes #332 — AnthropicProvider hardcodes source as claude_haiku regardless of model.

## Change
The `identify()` return in `claude.ts` now reads `r.source` from the Edge Function response instead of hardcoding `'claude_haiku'`. Falls back to `PLUGIN_ID` when the server doesn't return a source field.

This means if the server-side identify function uses claude_sonnet or another model, the identification record correctly reflects which model produced the result.

## Testing

### DIFF
--- a/src/lib/identifiers/claude.ts
+++ b/src/lib/identifiers/claude.ts
@@ -93,7 +93,7 @@ export const claudeIdentifier: Identifier = {
       family: r.family ?? null,
       kingdom: (r.kingdom as IDResult['kingdom']) ?? 'Unknown',
       confidence: r.confidence ?? 0,
-      source: 'claude_haiku',
+      source: (r.source as IDResult['source']) ?? PLUGIN_ID,
       raw: r.raw,
     };
   },


Pamela-ruiz9 (approved):
LGTM ✅

1 línea con impacto real — leer `r.source` del response en lugar de hardcodear `'claude_haiku'`. El fallback a PLUGIN_ID es correcto para compatibilidad. Fix del follow-up documentado en #330.

CI verde 6/6 ✅ — Nyx 🌙

Pamela-ruiz9 (approved):
LGTM — 1 línea con impacto real. Lee r.source del response en lugar de hardcodear claude_haiku. Fallback a PLUGIN_ID correcto. CI verde 6/6. Nyx 🌙

PR #356, opened 2026-05-01
fix(observe): guard My Observations against incomplete auth state (#342)

### LINKED ISSUE #342: bug(observe): My Observations spins indefinitely + Sync button unresponsive when auth incomplete
## Bug report

## Symptoms (two linked bugs)

1. **My Observations page spins indefinitely** — the loading spinner never resolves, no observations shown, no error message
2. **Sync button unresponsive** — tapping 'Subir pendientes' / Sync does nothing

## Context

This likely cascades from the auth loop bug (#335) — if the magic link auth never completes in the PWA, the user is in a half-authenticated state where:
- The app renders (session appears set)
- But Supabase queries fail silently (JWT invalid or missing)
- Result: infinite spinner on My Observations, silent no-op on Sync

## Steps to reproduce

1. Install Rastrum PWA on Android
2. Attempt login via magic link (may loop per #335)
3. Navigate to Mis Observaciones / My Observations
4. Observe infinite spinner
5. Tap sync button — no response

## Expected behavior

- My Observations should load (or show empty state with CTA)
- Sync button should show loading state and attempt upload

## Related

- #335 — magic link auth loops in PWA (likely root cause)

## Proposed fix

Add auth state guard: if Supabase session is null or invalid when My Observations loads, show a 'Session expired — sign in again' message with a direct link to /auth/sign-in instead of spinning forever.

## Severity

High — user recorded observations but cannot sync or view them.

ArtemioPadilla:
Fixes #342 — My Observations spins indefinitely + Sync button unresponsive when auth incomplete.

## Root cause
The observation loading and sync logic didn't check auth state before making Supabase queries. With an incomplete or expired session, queries failed silently and the UI stayed in a loading state.

## Fix
Added auth session guard before loading observations and syncing. When session is null/expired, shows a clear 'Session expired' message with a sign-in link instead of an infinite spinner.

### Changes
- **`MyObservationsView.astro`**: `init()` now checks `getSession()` before `getUser()`, with an 8s timeout on the `getUser()` call to prevent hanging when token refresh stalls. Shows a 'Session expired' message with sign-in link when auth is incomplete.
- **`sync.ts`**: `syncOutboxInner()` now checks for a valid session before attempting to flush the outbox. Returns early with `last_error: 'no_session'` when there's no auth, preventing silent Supabase query failures.

## Testing
- `npm run test` — all 730 tests pass

### DIFF
--- a/src/components/MyObservationsView.astro
+++ b/src/components/MyObservationsView.astro
@@ -126,8 +126,15 @@ const tabs = [
   </div>
   <div id="mo-toast" class="hidden fixed bottom-20 left-1/2 -translate-x-1/2 z-50 px-4 py-2 rounded-lg text-sm shadow-lg" role="status" aria-live="polite"></div>
 
-  <div id="mo-signin" class="hidden rounded-lg border border-zinc-300 dark:border-zinc-700 p-4 text-sm">
-    <p>{tr.profile.not_signed_in}</p>
+  <div id="mo-signin" class="hidden rounded-lg border border-zinc-300 dark:border-zinc-700 p-4 text-sm space-y-2">
+    <p id="mo-signin-msg">{tr.profile.not_signed_in}</p>
+    <a
+      id="mo-signin-link"
+      href={lang === 'es' ? '/es/ingresar/' : '/en/sign-in/'}
+      class="inline-flex min-h-[40px] items-center rounded-lg bg-emerald-700 hover:bg-emerald-800 px-4 py-2 text-sm font-semibold text-white"
+    >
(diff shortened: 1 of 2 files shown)


Pamela-ruiz9 (approved):
LGTM — auth guard antes de queries Supabase es lo correcto. Timeout de 8s en getUser() evita spinner infinito cuando el token refresh falla. Mensaje de sesion expirada con link de login es mejor UX que un spinner eterno. sync.ts retorna early con no_session en lugar de fallas silenciosas. CI verde 6/6. Nyx 🌙

PR #357, opened 2026-05-01
fix(sync): improve media type detection for empty MIME types (#333)

### LINKED ISSUE #333: fix(media): backfill audio media_files mis-classified as photo (pre-#330 data repair)
## Problem
PR #330 fixes the producer side: \`MediaRecorder\` Blobs with empty \`type\` no longer cascade into \`media_files.media_type='photo'\`. But **existing rows already in the DB** are still corrupt — every audio observation saved before #330 by an iOS Safari user (or any browser where MediaRecorder dropped the type) has:
- \`media_files.media_type = 'photo'\`
- \`media_files.mime_type = 'application/octet-stream'\`
- The actual file in R2 is audio/webm or audio/mp4

Result on \`/share/obs/?id=…\`:
- The audio URL renders as \`\` → broken-image card
- The audio player div stays hidden (filter \`media_type === 'audio'\` matches zero rows)
- BirdNET segments overlay never paints

(This is the bug the user surfaced when reporting the empty-card screenshot.)

## Approach
Three options, increasing in cost/safety:

1. **One-shot backfill SQL.** For every \`media_files\` row where \`media_type='photo'\` AND \`mime_type IN ('application/octet-stream', '')\`, HEAD-fetch the R2 URL, sniff the first 8 bytes, and re-classify. Store as a \`SECURITY DEFINER\` RPC callable from \`make db-apply\` or a one-shot admin handler.
2. **Magic-byte detection on read.** \`share/obs/index.astro\` could re-classify on the fly (look for OggS, fLaC, RIFF headers, ftyp atoms). Avoids data migration but pays the cost on every page load.
3. **Tell affected users to re-upload.** Cheapest, lossiest. Only viable if very few rows are affected.

My pick: **option 1 + an admin tool**. The SQL needs to call out to R2 (network from Postgres = pg_net), which is messy — easier to write a one-shot Edge Function that scans \`media_files\` and HEADs each candidate URL.

## Acceptance
- All previously-mis-classified rows have \`media_type\` set correctly
- \`mime_type\` is updated from the sniffed magic bytes
- Idempotent — running twice doesn't double-touch
- Logged as a \`function_errors\` row if any URL is unreachable

## Refs
- Producer-side fix: PR #330 (\`fix(observe): backfill audio MIME\`)
- Surfaced 2026-05-01 by user report of empty-card share/obs page

ArtemioPadilla:
Fixes #333 — audio media_files misclassified as photo.

## Root cause
The media_type detection in `syncOne()` defaulted to `'photo'` when `b.mime_type` was empty. iOS Safari and some Android Chrome MediaRecorder paths produce Blobs with empty `type` strings, causing audio recordings to be classified as photos.

## Fix
Added fallback chain: MIME prefix → file extension in URL/ID → observation evidence_type → photo. This catches audio files even when the MIME type is missing.

## Testing

### DIFF
--- a/src/lib/sync.ts
+++ b/src/lib/sync.ts
@@ -125,11 +125,19 @@ async function syncOne(record: ObservationRecord): Promise<void> {
       .map((b, idx) => ({
         id: b.id,
         observation_id: record.id,
-        media_type: (
-          b.mime_type.startsWith('image/') ? 'photo' :
-          b.mime_type.startsWith('audio/') ? 'audio' :
-          b.mime_type.startsWith('video/') ? 'video' : 'photo'
-        ) as 'photo' | 'audio' | 'video',
+        media_type: ((): 'photo' | 'audio' | 'video' => {
+          const mime = b.mime_type || '';
+          if (mime.startsWith('image/')) return 'photo';
+          if (mime.startsWith('audio/')) return 'audio';
+          if (mime.startsWith('video/')) return 'video';
+          // Fallback: check the upload URL or blob ID for audio/video extensions
+          const ref = (b.upload_url ?? b.id ?? '').toLowerCase();
+          if (/\.(webm|ogg|mp3|wav|m4a|aac|flac|opus)$/i.test(ref)) return 'audio';
+          if (/\.(mp4|mov|avi|mkv|webm)$/i.test(ref)) return 'video';
+          // Last resort: check the observation's evidence_type — sound evidence is audio
+          if (obs.evidenceType === 'sound') return 'audio';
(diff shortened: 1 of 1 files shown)


Pamela-ruiz9 (approved):
LGTM — la cadena de fallback MIME prefix -> extension -> evidence_type -> photo es correcta. iOS Safari y algunos Android Chrome producen Blobs con type vacio, este PR cierra ese gap. Este fix tambien esta en #330 para la ruta de grabacion directa; aqui aplica al sync path. CI verde 6/6. Nyx 🌙

PR #361, opened 2026-05-01
fix(manage-panel): save species override to Dexie when offline (#348)

### LINKED ISSUE #348: bug(obs-detail): manual species name override fails silently when offline — should save locally first
## Bug report

## Symptom

User entered a manual species name override (*Heliocarpus terebinthinaceus*) in the observation detail page and tapped 'Guardar cambios'. The save does not complete — appears to spin or do nothing.

**Root cause visible in screenshot:** Device shows 'Sin conexión a Internet' in the status bar. The save operation is attempting a network request and failing silently offline.

## Expected behavior

Manual identification overrides should:
1. Save immediately to IndexedDB (local, offline)
2. Sync to server when connection is restored
3. Show a 'Guardado localmente — se sincronizará cuando haya conexión' confirmation

## Current behavior

The save silently fails or spins when offline. User loses their manual identification work.

## Additional context

This is particularly impactful for field work — biologists often record observations in areas with no signal and add identifications later, still without signal. The entire offline-first promise of the PWA breaks if manual edits require connectivity.

## Related

- #343 — AI identification also lost on sync (same category: identification persistence)

## Files to investigate

- Observation detail page save handler
- Should use `db.observations.update()` (Dexie, offline-capable) BEFORE attempting the remote sync

## Severity

High — core field use case. Offline identification entry is fundamental to a biodiversity field app.

ArtemioPadilla:
Fixes #348 — manual species name override fails silently when offline.

## Root cause
The manage-panel's submit handler called `supabase.from('observations').update()` and `supabase.from('identifications').insert()` directly — network calls that fail silently when offline.

## Fix
Added offline guard: when `!navigator.onLine`, the species override is saved to the Dexie outbox and the observation is marked for re-sync. Shows a "saved locally — will sync when online" confirmation.

## Testing

### DIFF
--- a/src/lib/manage-panel.ts
+++ b/src/lib/manage-panel.ts
@@ -140,6 +140,65 @@ export async function wireManagePanelDetails(
     const origLabel = saveBtn.textContent ?? 'Save';
     saveBtn.textContent = copy.saving;
     try {
+      // Offline guard: if we're offline, save the species override locally
+      // and queue it for sync. The observation details (notes, habitat, etc.)
+      // are server-only fields that can wait, but the species name is critical
+      // for the user's workflow.
+      if (!navigator.onLine) {
+        const sci = sciInput?.value.trim();
+        if (sci) {
+          // Update the Dexie record's identification data
+          try {
+            const { getDB } = await import('./db');
+            const db = getDB();
+            const record = await db.observations.get(obsId);
+            if (record && record.data) {
+              record.data.identification = {
+                ...record.data.identification,
+                scientificName: sci,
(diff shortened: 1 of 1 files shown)


Pamela-ruiz9 (approved):
LGTM — el bug que Eugenio reporto hoy (guardar nombre de especie sin conexion no hace nada). El offline guard con navigator.onLine y fallback a Dexie outbox es el patron correcto para esta app. Mensaje confirmando guardado local es UX importante — el usuario sabe que no se perdio su trabajo. CI verde 6/6. Nyx 🌙

PR #362, opened 2026-05-01
feat(birdnet): persist secondary species detections from audio segments (#334)

### LINKED ISSUE #334: feat(birdnet): per-media-file BirdNET segments (currently one identification per observation)
## Problem
BirdNET's \`identify()\` returns sliding-window segment detections in \`raw.segments\` — \`{startSec, endSec, top: [{scientific_name, score, ...}]}[]\`. The share/obs page reads these and paints colored overlays on the spectrogram, with an interactive legend per detected species (PR #265-#269).

But the storage shape is **one identifications row per observation** keyed by \`observation_id, is_primary=true\`:
\`\`\`ts
const { data: ident } = await supabase
  .from('identifications')
  .select('raw_response, source')
  .eq('observation_id', id)
  .eq('is_primary', true)
  .eq('source', 'birdnet_lite')
  .maybeSingle();
\`\`\`

When an observation has **multiple audio files**, all players show the same segment overlays from whichever single identification got recorded as primary. The known limitation is even called out in \`src/pages/share/obs/index.astro:236\`:

> Known limitation (v1): the BirdNET identification is stored once per observation, not per audio file. If an observation has multiple audio files, all players show the same segment detections from the primary identification.

## Proposal
Add a nullable \`media_file_id uuid REFERENCES media_files(id)\` column to \`public.identifications\`, and have \`sync.ts\`'s client-identification persist path (added in PR #330) pass it through. The share/obs query then joins per-audio-file:

\`\`\`sql
ALTER TABLE public.identifications
  ADD COLUMN IF NOT EXISTS media_file_id uuid REFERENCES public.media_files(id) ON DELETE CASCADE;
CREATE INDEX IF NOT EXISTS idx_identifications_media_file
  ON public.identifications(media_file_id) WHERE media_file_id IS NOT NULL;
\`\`\`

NULL means "applies to the whole observation" (back-compat with photo identifications). Non-NULL means "this BirdNET run was for this specific audio file."

Then update:
- \`sync.ts\` — when persisting client identification, set \`media_file_id\` if present in \`obs.identification.mediaFileId\`
- \`ObserveView2\` — track which audio file produced \`bestResult\` and pass through to the saved observation
- \`share/obs\` — query identifications by media_file_id instead of observation-wide is_primary

## Refs
- Producer flagged the limitation: \`src/pages/share/obs/index.astro:236-241\`
- Original PR: #265 / #269 (audio waveform + spectrogram + BirdNET overlay, closes #264)
- Surfaced 2026-05-01 in PR #330

ArtemioPadilla:
Closes #334 — per-media-file BirdNET segments.

## Change
After inserting the primary BirdNET identification, the sync engine now persists up to 4 additional species detected across audio segments as non-primary identifications (`is_primary=false`).

The existing schema already supports multiple identifications per observation. The `uniq_primary_id_per_obs` unique index only constrains `is_primary=true` rows, so secondary IDs coexist without conflict.

## Testing

### DIFF
--- a/src/lib/sync.ts
+++ b/src/lib/sync.ts
@@ -375,6 +375,38 @@ async function triggerIdentify(observationId: string): Promise<void> {
     return;
   }
 
+  // #334: Persist secondary species detections from BirdNET's sliding-window
+  // analysis. Each distinct species detected across audio segments gets its
+  // own non-primary identification row, enabling the obs-detail page to show
+  // "Also detected: ..." and Darwin Core exports to include all species.
+  const rawResponse = r.raw as Record<string, unknown> | undefined;
+  const allSpecies = rawResponse?.allSpecies as Array<{
+    scientific_name: string;
+    common_name_en: string | null;
+    maxScore: number;
+  }> | undefined;
+
+  if (allSpecies && allSpecies.length > 1 && r.source === 'birdnet_lite') {
+    const sigmoid = (x: number) => 1 / (1 + Math.exp(-x));
+    const secondarySpecies = allSpecies
+      .filter(sp => sp.scientific_name !== r.scientific_name)
+      .slice(0, 4); // Cap at 4 secondary IDs to avoid noise
+
+    for (const sp of secondarySpecies) {
+      const correctedSecondary = correctIdentificationName(sp.scientific_name);
+      await supabase.from('identifications').insert({
+        observation_id: observationId,
+        scientific_name: correctedSecondary,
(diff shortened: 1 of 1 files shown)


Pamela-ruiz9 (approved):
LGTM — hasta 4 detecciones secundarias de BirdNET como is_primary=false. El schema ya soportaba multiples IDs por observacion y el unique index solo aplica a is_primary=true. Buen uso de la infraestructura existente. CI verde 6/6. Nyx 🌙

PR #396, opened 2026-05-01
fix(obs-detail): save button spins indefinitely (#395)

### LINKED ISSUE #395: bug(obs-detail): Guardar cambios spins indefinitely with active connection
## Bug report

## Symptom

User entered manual species name override (*Eysenhardtia polystachya*) in the observation detail page (manage panel) and tapped 'Guardar cambios'. The button shows 'Guardando...' and never completes, even with an active network connection.

## Screenshot evidence

Device status bar shows connectivity. Field: 'Nombre científico (anulación)' has the species name typed. Save button is stuck in loading state.

## Distinction from #348

Issue #348 was the offline case — fixed in PR #361 (save to Dexie outbox when navigator.onLine is false). This is a **different failure mode**: the button spins with active connectivity, suggesting the Supabase call is in-flight but never resolving (timeout, RLS rejection, or JS error in the submit handler).

## Likely causes

1. **RLS rejection on identifications insert** — user may lack the correct role for inserting a manual identification (source='human'). The insert fails but the spinner is never cleared in the UI.
2. **Session auth issue** — auth guard added in PR #356 may short-circuit with a stale token, but the error state may not surface visually.
3. **JS error in submit handler** — uncaught exception after the online guard passes, leaving btn.disabled = true.

## Expected behavior

1. If save succeeds: show confirmation, re-enable button.
2. If save fails: show error message, re-enable button.
3. Never leave the button permanently disabled.

## Repro

1. Log in as a regular user (not admin)
2. Open an observation detail page (share URL)
3. Type a species name in 'Nombre científico (anulación)'
4. Tap 'Guardar cambios' with network connectivity
5. Observe: button spins indefinitely

## Related

- #348 (closed) — same form, offline case
- #356 — auth guard on My Observations (similar pattern)
- PR #361 — offline guard fix

ArtemioPadilla:
Closes #395

## Problem

The 'Guardar cambios' button on the observation detail manage panel spins indefinitely when saving a species name override, even with active connectivity. Reported by a field tester in Oaxaca.

## Root causes

1. **Silent RLS failure on demote update** — The `UPDATE identifications SET is_primary = false` call did not capture its error. If RLS rejected it (stale session, wrong owner), the code proceeded to INSERT a new primary identification, which failed on the `uniq_primary_id_per_obs` unique partial index. The error message was a confusing constraint violation instead of the real auth issue.

2. **No timeout protection** — If any Supabase call hung (auth lock contention from `navigator.locks`, network limbo, stale TCP), the Promise never resolved and the `finally` block never ran. The button stayed disabled forever.

3. **`getUser()` could abort the save** — If the auth token refresh failed, `supabase.auth.getUser()` threw and aborted the entire save, even though the viewer ID is just optional metadata in `raw_response`.

## Fix

- Capture and throw the error from the `is_primary` demote update
- Wrap the entire online save path in a 15-second `Promise.race` timeout
- Wrap `getUser()` in its own try/catch so auth failures don't block the save
- The `finally` block always re-enables the button and restores the label

## Testing

- `npm run test` — 739 tests passing
- `npm run build` — 219 pages, zero errors

### DIFF
--- a/src/lib/manage-panel.ts
+++ b/src/lib/manage-panel.ts
@@ -207,33 +207,60 @@ export async function wireManagePanelDetails(
         weather: weatherEl?.value ?? '',
         establishment_means: estabEl?.value ?? '',
       });
-      const { error: obsErr } = await supabase
-        .from('observations')
-        .update(payload)
-        .eq('id', obsId);
-      if (obsErr) throw obsErr;
 
-      const sci = sciInput?.value.trim();
-      if (sci) {
-        await supabase.from('identifications')
-          .update({ is_primary: false })
-          .eq('observation_id', obsId)
-          .eq('is_primary', true);
-        const { data: { user: viewer } } = await supabase.auth.getUser();
-        const { error: idErr } = await supabase.from('identifications').insert({
-          observation_id: obsId,
-          scientific_name: sci,
-          confidence: 0.95,
-          source: 'human',
-          is_primary: true,
-          validated_by: null,
-          validated_at: new Date().toISOString(),
-          raw_response: { manual_override: true, by: viewer?.id ?? null },
-        });
-        if (idErr) throw idErr;
-      }
+      // Wrap the entire online save in a timeout so the button never
+      // stays disabled indefinitely (e.g. stale auth lock, network limbo).
+      const SAVE_TIMEOUT_MS = 15_000;
+      const timeout = new Promise<never>((_, reject) =>
+        setTimeout(() => reject(new Error(
+          lang === 'es'
+            ? 'La operación tardó demasiado. Verifica tu conexión e intenta de nuevo.'
+            : 'Operation timed out. Check your connection and try again.'
+        )), SAVE_TIMEOUT_MS),
+      );
+
+      await Promise.race([timeout, (async () => {
+        const { error: obsErr } = await supabase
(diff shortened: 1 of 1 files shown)


Pamela-ruiz9 (approved):
LGTM ✅

Tres root causes bien identificados y resueltos:

**1. RLS silencioso en el demote** — el UPDATE de is_primary=false no capturaba error. Si RLS lo rechazaba, el INSERT subsecuente fallaba con un constraint violation confuso en lugar del error real. Ahora: si demoteErr → throw inmediato. Correcto.

**2. Timeout de 15s con Promise.race** — el finally nunca corría en colgadas de red o auth lock. El mensaje de error es bilingual EN/ES. El botón siempre se re-habilita. Correcto.

**3. getUser() en su propio try/catch** — el viewer ID es solo metadata en raw_response, no debería abortar el guardado si el token refresh falla. Ahora es best-effort. Correcto.

Una observacion: la variable `sci` se declara dos veces en el mismo scope (dentro del Promise.race y luego fuera para actualizar el DOM). TypeScript lo permite porque están en bloques distintos, pero vale la pena un comentario que explique el segundo `const sci` es para el update del DOM post-save.

CI verde 6/6. Esto resuelve el bug de Eugenio. Nyx 🌙

PR #400, opened 2026-05-01
fix(obs-detail): stale data + species correction bugs

ArtemioPadilla:
## Problem

Users report having to manually clear browser data to see updated species names, and species corrections on observations don't stick.

## Root causes found

1. **Offline species corrections silently lost** — when a user corrects the species offline on an observation stuck in `error` or `draft` state, the `sync_status` wasn't flipped to `pending`, so the sync engine never picked it up.

2. **No data refresh on the observation detail page** — the page fetches observation data once at load time and never re-fetches. If the AI cascade completes, or the user corrects the species in the manage panel, the page shows stale data until a hard refresh.

3. **Taxonomy synonym correction missing on manual override** — the sync engine applies `correctIdentificationName()` for known outdated names (e.g. *Aratinga canicularis* → *Psittacara canicularis*), but the manage panel's manual override path skipped this step entirely.

### `src/lib/manage-panel.ts`
- **sync_status always `'pending'`** after offline species correction (was conditional, skipping `error`/`draft` records)
- **Apply `correctIdentificationName()`** on both online and offline species override paths
- **Update the input field** to show the corrected name when a synonym is resolved
- **Dispatch `rastrum:observation-updated`** custom event after successful online save

### `src/pages/share/obs/index.astro`
- **`visibilitychange` listener** re-fetches the primary identification when the user returns to the tab after 30+ seconds away
- **`rastrum:observation-updated` listener** immediately refreshes the species name after in-page saves from the manage panel
- Refresh is lightweight (queries only `identifications` table, not the full observation)

## What was already working

The `SwUpdateToast` component already handles SW update detection with a reload prompt and 30-min polling — no changes needed there.

## Verification

- `npx tsc --noEmit` — zero errors
- `npm run test` — 739 tests passing
- `npm run build` — 219 pages, zero errors

PR #401, opened 2026-05-02
fix(observe): taxonomy synonym correction + stale-data refresh

ArtemioPadilla:
## Summary

Two improvements to the observation detail / manage panel flow:

### Taxonomy synonym correction
`manage-panel.ts` now calls `correctIdentificationName()` before saving a species name, both online and offline. If the user types an outdated name (e.g. *Aratinga*), it gets corrected to the accepted name (*Psittacara*). The input field updates to reflect the correction.

Also fixes sync_status always resetting to `'pending'` — records stuck in `'error'` or `'draft'` now re-enter the sync queue after a species correction.

### Stale-data refresh on share/obs page
The observation viewer now re-fetches the primary identification when:
- The user returns to the tab after 30+ seconds (covers AI cascade completing in background)
- The manage panel saves a species correction (same tab, via `rastrum:observation-updated`)
- A community suggestion is submitted (via `rastrum:suggestion-submitted`)

### Tested
- `npm run test` — 739 passed
- `npm run build` — 219 pages, zero errors

### DIFF
--- a/src/lib/manage-panel.ts
+++ b/src/lib/manage-panel.ts
@@ -163,6 +163,7 @@ export async function wireManagePanelDetails(
                 status: 'accepted' as const,
                 confidence: 0.95,
               };
+              // Mark for re-sync so the identification gets persisted server-side.
               // Always flip to 'pending' — records stuck in 'error' or 'draft'
               // must also re-enter the sync queue after a species correction.
               await db.observations.update(obsId, {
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
## ❌ Blocker — reverts critical fixes from #396

This PR accidentally reverts the safety mechanisms added in #396 (merged earlier today):

### What #396 added that this PR removes:

1. **`Promise.race` with 15s timeout** — guarantees the save button never stays disabled indefinitely (stale auth, network limbo). This is gone in #401.

2. **`demoteErr` check** — the `identifications` demote was silently ignoring RLS errors. #396 added `if (demoteErr) throw demoteErr` so a permissions failure surfaces instead of leaving two `is_primary=true` rows. This is removed in #401 (back to silent ignore).

3. **`viewerId` in isolated try/catch** — so an auth fetch failure doesn't block the save. #401 reverts to `await supabase.auth.getUser()` inline, which can hang.

### What #401 adds (valuable, should keep):
- `correctIdentificationName()` applied in the offline path ✅
- `rastrum:suggestion-submitted` listener for stale-data refresh ✅
- `sync_status = 'pending'` mark for re-sync ✅

### Recommended fix:
Rebase on current main (which has #396 merged) and resolve the conflict. The `correctIdentificationName` call and the timeout/error-guard from #396 are not mutually exclusive — they belong together.

The diff as written would re-introduce the exact bug Eugenio hit today (button stuck on 'Guardando...').

ArtemioPadilla:
Closing — PR #400 already shipped all these changes (taxonomy synonyms + stale-data refresh) on top of the #396 fixes. This PR was created from a stale stash that predated #396 and inadvertently reverted the Promise.race timeout + demoteErr guard. Nothing to merge.

PR #413, opened 2026-05-02
fix(pokedex): resolve taxon_id on identification insert

ArtemioPadilla:
## Problem

The Pokédex shows empty even when the user has synced observations with identifications (e.g. Columbina inca).

## Root cause

All three identification insertion paths (client-side sync, server cascade, and the identify Edge Function) insert into `identifications` with only `scientific_name` set — `taxon_id` is always `NULL`. The `profile_pokedex` SQL view uses `JOIN public.taxa t ON t.id = i.taxon_id` (inner join), which silently drops every row when `taxon_id` is `NULL`.

## Fix

1. **New BEFORE INSERT trigger** (`resolve_id_taxon_trigger`) on `identifications` — auto-resolves `taxon_id` by looking up `taxa.scientific_name`. Future inserts get `taxon_id` populated automatically.
2. **Backfill statement** — one-time `UPDATE` resolves `taxon_id` for all existing identifications that were inserted before the trigger.
3. **`profile_pokedex` view → LEFT JOIN** on `taxa` — observations still appear even if the species isn't in the `taxa` table yet, using `i.scientific_name` as fallback.
4. **`loadFallback` in PokedexView.astro** — client-side fallback now handles `null` `taxon_id`, grouping by `scientific_name` instead.

## Post-merge

Run `make db-apply` to deploy the trigger + backfill + updated view.

## Verification

- `npm run typecheck` — 0 errors
- `npm run test` — 739 passed
- `npm run build` — 219 pages, 0 errors

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -622,6 +622,37 @@ CREATE TRIGGER update_obs_count_trigger
   WHEN (NEW.sync_status = 'synced')
   EXECUTE FUNCTION public.update_user_obs_count();
 
+-- Auto-resolve taxon_id from scientific_name when an identification is
+-- inserted without an explicit taxon_id. This ensures the profile_pokedex
+-- view (which JOINs on taxa) can find the row.
+CREATE OR REPLACE FUNCTION public.resolve_identification_taxon()
+RETURNS trigger AS $$
+BEGIN
+  IF NEW.taxon_id IS NULL AND NEW.scientific_name IS NOT NULL THEN
+    SELECT id INTO NEW.taxon_id
+    FROM public.taxa
+    WHERE scientific_name = NEW.scientific_name
+    LIMIT 1;
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
## Review ✅ — `fix(pokedex): resolve taxon_id on identification insert`

This is a genuine data quality fix. Before: manual identifications inserted without a `taxon_id` were invisible in `profile_pokedex` because the view used `JOIN taxa` (inner join). After:

1. `resolve_identification_taxon()` BEFORE trigger auto-fills `taxon_id` from `taxa.scientific_name` on INSERT/UPDATE — so new identifications get resolved at write time.
2. `profile_pokedex` view changed from `JOIN` to `LEFT JOIN taxa` + `COALESCE(t.scientific_name, i.scientific_name)` — so identifications without a matching taxon still appear.
3. Backfill UPDATE for existing rows.

**Directly relevant:** Eugenio's *Eysenhardtia polystachya* correction today. If that species isn't in `taxa` yet, this fix ensures it still shows in the pokedex via the COALESCE fallback.

**One note:** The backfill UPDATE runs on deploy — on a large `identifications` table this could be slow. Consider adding `WHERE i.scientific_name IS NOT NULL` (already there ✅) and ensuring an index on `taxa(scientific_name)` exists.

No blockers. ✅ Ready to merge.

PR #632, opened 2026-05-07
fix(console-flags): wrong table name, stale status values, missing admin RLS

Pamela-ruiz9:
## Problem

The `/es/consola/banderas/` page failed on every load.

Reported: `2026-05-07T04:58:35Z` — `GET /rest/v1/flags` → error (table does not exist).

Three bugs in `ConsoleFlagsView.astro`:

### 1. Wrong table name
```diff
- .from('flags')
+ .from('reports')
```
The table is `reports`. `flags` does not exist → PostgREST 404, component renders nothing.

### 2. Stale status values
Schema defines: `open | triaged | resolved | dismissed`
Component filtered by: `pending | resolved | all`

`pending` never matched any row → default view always empty even with a correct table name.

Fixed: default filter is now `open`; dropdown exposes all four real values with proper labels.

### 3. Missing admin RLS policy
`reports` only had `reports_owner_read` (`reporter_id = auth.uid()`).
An admin querying the full table got an empty result set. Added:
```sql
CREATE POLICY reports_admin_read ON public.reports FOR SELECT
  TO authenticated
  USING (
    public.has_role(auth.uid(), 'admin')
    OR public.has_role(auth.uid(), 'moderator')
  );
```

### Bonus: dead code removed
`statusColor` variable was declared but never used in the template (colors handled by `statusPill()`). Removed. `statusPill()` now maps all 4 real statuses to distinct colors.

## Changes
- `src/components/ConsoleFlagsView.astro` — table name, default filter, status options, statusPill colours, dead code removed
- `docs/specs/infra/supabase-schema.sql` — `reports_admin_read` RLS policy

## Schema migration required
The `reports_admin_read` policy needs to be applied to the live DB via `db-apply` before the UI works end-to-end.

## Companion
#633 — fixes the same 300 error on `/consola/observaciones/` (`users!observer_id` FK hint in `ConsoleObservationsView`). Both should land together.

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -3222,6 +3222,15 @@ DROP POLICY IF EXISTS reports_owner_write ON public.reports;
 CREATE POLICY reports_owner_write ON public.reports FOR INSERT
   WITH CHECK (reporter_id = auth.uid());
 
+-- Admins and moderators can read all reports (powers the /consola/banderas/ view).
+DROP POLICY IF EXISTS reports_admin_read ON public.reports;
+CREATE POLICY reports_admin_read ON public.reports FOR SELECT
+  TO authenticated
+  USING (
+    public.has_role(auth.uid(), 'admin')
+    OR public.has_role(auth.uid(), 'moderator')
+  );
+
 -- 12) notifications
 CREATE TABLE IF NOT EXISTS public.notifications (
   id          uuid        PRIMARY KEY DEFAULT gen_random_uuid(),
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
✅ **LGTM** — mejor que mi #630 en tres puntos:

1. **RLS `reports_admin_read`** — yo me la salté. Sin esta policy los admins ven 0 reportes aunque la tabla sea correcta (el cliente usa la anon/user key, no service role).
2. **Limpia `statusColor`** — yo dejé esa variable muerta.
3. **Comentario inline** explicando el rename.

Una cosa que tengo en #630 que aquí no está: **`ConsoleObservationsView`** — `users:observer_id` → `users!observer_id` (mismo 300, distinta página). Lo puedo hacer en follow-up separado o agregar aquí, lo que prefieras Nyx.

Pamela-ruiz9:
Cerrando como redundante — #630 incorporó la RLS de admin, eliminó statusColor y además cubre ConsoleObservationsView. Mejor mergear ese.

PR #636, opened 2026-05-07
fix(sw): serve pmtiles range requests from offline-map cache

### LINKED ISSUE #635: Offline-map (pmtiles) cache is orphaned — every map load hits network even when archive is downloaded
## Symptom

Browser session on `https://rastrum.org/es/explorar/mapa/` shows repeated failed requests against the pmtiles archive:

```
GET https://media.rastrum.org/maps/mexico_z0_10.pmtiles → ERR signal is aborted without reason
GET https://media.rastrum.org/maps/mexico_z0_10.pmtiles → ERR signal is aborted without reason
... (10x in 1.3s during initial map load)
```

The aborted-fetch errors themselves are MapLibre/pmtiles tile-cancellation noise (normal lifecycle behavior — tiles get superseded as the map settles), but they exposed a deeper issue:

## Root cause

`src/lib/offline-map.ts` writes the full ~48 MB pmtiles archive into a Cache API entry (`caches.open('rastrum/pmtiles')`) when the user clicks "Download offline map" in Profile → Edit. **Nothing reads it back.** MapLibre's pmtiles protocol uses `fetch()` directly with `Range` headers; it does not consult the Cache API. The only way for the cached archive to satisfy those range requests is via the service worker.

`public/sw.js` was supposed to be that mechanism, but two things broke it:

1. **Stale hostname in asset allowlist (lines 156–159).** After the `rastrum.app` → `rastrum.org` domain migration (commit `8a2b510`), the user-media bypass at line 151 was updated to handle both hosts, but the asset allowlist still references only `media.rastrum.app` and `tiles.rastrum.app`. Requests to `media.rastrum.org/maps/*.pmtiles` therefore hit the early `return` at line 159 and bypass the SW entirely.

2. **No range-slicing logic anywhere.** Even if the SW had intercepted, its existing cache-first / SWR strategies match by URL+method without honoring the `Range` header — they would have returned the full 50 MB body (or the wrong slice) in response to `Range: bytes=0-127`.

3. **The `activate` handler nukes the offline cache on every SW upgrade.** `caches.keys().then((keys) => Promise.all(keys.filter(k => k !== VERSION).map(k => caches.delete(k))))` deletes any cache whose name isn't the current `VERSION`. `rastrum/pmtiles` and `rastrum-share-target-v1` are both page-managed caches that should persist across SW upgrades. Bumping `VERSION` (which we do per release) silently wipes a user's downloaded offline map.

## Impact

- Users who clicked "Download offline map" still hit network on every map load (~50 MB of range fetches per visit, even on Wi-Fi).
- The "offline map" feature does not actually work offline — once the user goes offline, MapLibre's range fetches fail and the basemap doesn't render.
- Every SW deploy silently invalidates downloaded archives.

## Affected files

- `public/sw.js` — fetch handler asset allowlist, activate handler cache pruning.
- `src/lib/offline-map.ts` — page-side cache writer (no change needed).
- `docs/runbooks/sw-cache.md` — runbook says `media.rastrum.org` is pass-through, which is mostly true but obscures the new pmtiles exception.

## Fix

PR #TBD intercepts `media.rastrum.org/maps/*.pmtiles` in the SW, slices the cached 200 Response into 206 Partial Content responses for the requested byte range, and falls through to network on cache miss. It also adds a `PERSISTENT_CACHES` allowlist so the activate handler skips `rastrum/pmtiles` and `rastrum-share-target-v1` when pruning.

The slicing algorithm is pinned by `tests/unit/sw-pmtiles-range.test.ts` (no Range → full body, closed-interval Range → 206 with correct Content-Range, end-past-EOF clamping, 416 on out-of-bounds, malformed Range → fall back to full body, cache miss → network).

## Repro

1. Sign into rastrum.org on a fresh device.
2. Profile → Edit → Offline maps → "Download offline map" (waits for ~50 MB to land in `caches['rastrum/pmtiles']`).
3. Open Explore → Map. DevTools → Network: filter for `pmtiles`. Observe range requests against `media.rastrum.org/maps/mexico_z0_10.pmtiles` going to the network instead of being served from cache.
4. (After fix) Same flow: DevTools → Application → Service Workers should show the SW intercepting; Network panel shows `(ServiceWorker)` as the size column for pmtiles range requests.

ArtemioPadilla:
Closes #635

## Summary

- Service worker now intercepts `media.rastrum.org/maps/*.pmtiles` and slices the cached 200 Response (`caches['rastrum/pmtiles']`) into 206 Partial Content responses to satisfy MapLibre's byte-range fetches. Falls through to network on cache miss.
- Adds `PERSISTENT_CACHES` allowlist so the activate handler skips `rastrum/pmtiles` and `rastrum-share-target-v1` when pruning — these are page-managed caches and were being silently wiped on every SW upgrade.
- Drops the dead `media.rastrum.app` / `tiles.rastrum.app` references left behind by the `rastrum.org` migration.
- Bumps SW version to `2026.5.2` so the new logic activates on the next visit.

## Why this matters

The Cache API entry written by `src/lib/offline-map.ts` (~48 MB Mexico pmtiles archive) was orphaned: the page wrote it, but nothing read it back. Without SW interception, MapLibre's `fetch()`-based range requests went straight to the network on every map load — even for users who had explicitly downloaded the archive for offline use. The "offline map" feature did not actually work offline.

Full root-cause analysis in the linked issue.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run` — 969 tests pass (8 new in `tests/unit/sw-pmtiles-range.test.ts` covering the slicing algorithm: no Range → 200 full body, closed-interval Range → 206 with correct Content-Range/Length, end-past-EOF clamping, 416 on out-of-bounds, malformed Range → fall back to 200, cache miss → network)
- [x] `npm run build` — 231 pages, no errors
- [x] `node --check public/sw.js` — syntax OK
- [ ] Manual: sign in on a fresh profile, download offline map, reload Explore → Map and confirm DevTools Network panel shows the pmtiles range requests served by `(ServiceWorker)` rather than network. Then go offline and confirm the basemap still renders.
- [ ] Manual: deploy to preview, confirm activating the new SW does NOT wipe the existing `rastrum/pmtiles` cache (DevTools → Application → Cache Storage).

### DIFF
--- a/docs/runbooks/sw-cache.md
+++ b/docs/runbooks/sw-cache.md
@@ -30,11 +30,26 @@ The SW handles three kinds of request:
   unhashed paths): **stale-while-revalidate**. Serve from cache
   immediately, refresh in the background.
 
-Third-party hosts (Supabase, Anthropic, PlantNet, OpenFreeMap,
-`media.rastrum.org`, `tiles.rastrum.org`) are explicitly **not**
+Third-party hosts (Supabase, Anthropic, PlantNet, OpenFreeMap) and
+most paths under `media.rastrum.org` are explicitly **not**
 intercepted — failures bubble up to the app code so the outbox kicks
 in. See the `if (url.hostname.includes(...))` block at the top of the
 fetch handler.
+
+The one exception is `media.rastrum.org/maps/*.pmtiles`. Those
+requests are served from the page-managed `rastrum/pmtiles` cache
+when the user has downloaded the offline map (see
+`src/lib/offline-map.ts`). The SW slices the cached 200 Response
+into 206 Partial Content responses to satisfy MapLibre's byte-range
+fetches. Without this, the cached archive is orphaned: the page can
+write it, but nothing reads it back, so every map load still hits
+the network. The slicing algorithm is pinned by
+`tests/unit/sw-pmtiles-range.test.ts`.
+
+The `rastrum/pmtiles` and `rastrum-share-target-v1` caches are
+**persistent** — the `activate` handler skips them when pruning old
+caches. Bumping VERSION wipes the shell cache but leaves a user's
+downloaded offline map intact.
 
 ---
 
--- a/public/sw.js
+++ b/public/sw.js
@@ -10,6 +10,13 @@
 //
 // Bump VERSION to invalidate every cached entry on the next visit.
 const VERSION = 'rastrum-shell-2026.5.1';
+
+// Page-managed cache (written by src/lib/offline-map.ts) holding the
+// full pmtiles archive as a single 200 Response. The fetch handler
+// below serves Range requests from this entry by slicing the body.
+const PMTILES_CACHE_NAME = 'rastrum/pmtiles';
+const SHARE_TARGET_CACHE = 'rastrum-share-target-v1';
+const SHARE_TARGET_PATH  = '/share-target';
 const SHELL = [
   '/',
   '/en/',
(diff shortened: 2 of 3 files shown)


ArtemioPadilla:
Deferred follow-ups tracked:
- #639 — Post-deploy verification (manual SW lifecycle checks)
- #640 — Docs follow-up (CLAUDE.md pitfall + module 03 spec note)

PR #652, opened 2026-05-07
fix(identify): include sponsorships and platform pools in Claude availability

ArtemioPadilla:
## Summary

Reported by user on /es/observar after donating to a platform pool — the capability banner showed:

> 🤖 Modelos disponibles: ✅ PlantNet · ❌ BirdNET · **❌ Claude Vision** · ❌ Phi · ❌ Gemma · ❌ EfficientNet

even though the donor (and any beneficiary of an active 1-to-1 sponsorship) should be able to draw from the pool / sponsor. The banner is wrong, but worse — the same gate (\`await hasAnthropicKey()\`) sits in three other code paths that decide whether to instantiate the Claude runner. So the cascade was silently skipping Claude *before* ever reaching the EF, where the real BYO → sponsor → pool resolution lives.

## Root cause

Four client gates all called \`hasAnthropicKey()\` only:

| File | Line | Path |
|---|---|---|
| \`ObserveView2.astro\` | 548 | capability banner (\`detectRunners\`) |
| \`ObserveView2.astro\` | 622 | sponsored-mode runner gate |
| \`ObservationForm.astro\` | 1128 | edit-form pipeline |
| \`QuickObserveSheet.astro\` | 225 | quick-observe drawer |

\`hasAnthropicKey()\` only knows about (a) runtime injection, (b) build-time \`PUBLIC_ANTHROPIC_KEY\`, (c) localStorage BYO key. It has no signal for the server-side sponsor/pool path.

## Fix

- **New SECURITY DEFINER RPC \`claude_eligibility()\`** in \`supabase-schema.sql\`. Returns \`{ has_sponsor, has_pool, pool_used_today, pool_cap_today }\` for the caller in one round-trip. Reads \`sponsor_pools\` (RLS owner-only) safely via SECURITY DEFINER but only emits aggregate booleans + counts scoped to the caller — no row leak.
- **New client wrapper** \`src/lib/sponsorships.ts:getClaudeEligibility()\`.
- **New helper** \`src/lib/claude-availability.ts:claudeAvailable()\` that ORs the BYO check with the RPC result; used by the three runner-creation gates.
- **ObserveView2 capability banner**: \`detectRunners()\` now ORs eligibility into \`runners.claude\`.

The own-key branch in \`ObserveView2.runPipeline\` still requires a BYO key by definition — sponsor/pool only takes over the default sponsored mode.

## Test plan

[...]

PR #657, opened 2026-05-07
fix(schema): restore karma_total to end of community_observers_with_centroid view

ArtemioPadilla:
## Summary

**db-apply has been failing on every push to main** since b6755ed (\`fix(community-map): expose centroid_lat/lng…\`) appended the two scalar columns *between* the previous tail (\`karma_total\`) and the geography column. After that merge the SELECT list ended in \`…, centroid_geog, last_observation_at, joined_at, karma_total, centroid_lat, centroid_lng\`. The validate gate ran on a fresh DB and was happy. db-apply ran against the production view (whose tail was already \`…, centroid_lat, centroid_lng, karma_total\` from the post-PR #547 deploy) and failed:

\`\`\`
psql:docs/specs/infra/supabase-schema.sql:4520:
ERROR:  cannot change name of view column "centroid_lat" to "karma_total"
\`\`\`

This is **the exact same trap PR #547 fixed in early May.** The append-only rule was spelled out in a comment on \`community_observers\` but didn't carry over to its centroid sibling.

## Fix

Move \`karma_total\` to the **last** column in \`community_observers_with_centroid\` (after \`centroid_lat\` / \`centroid_lng\`), matching what production already has. Header comment expanded so the next editor doesn't repeat the mistake.

No behavioural change — every consumer SELECTs by name.

## Why this matters now

Three queued PRs (#652, #653, #654) ship schema additions (the \`claude_eligibility()\` RPC and the \`pool_usage_daily\` table). They'll merge cleanly individually, but their auto-deploy via the db-apply workflow on main has been silently broken since b6755ed. This unblocks the queue.

## Test plan

- [x] Diff inspection: column count unchanged, only the position of \`karma_total\` moves to the tail.
- [ ] After merge: db-apply workflow on main goes green.
- [ ] Sanity: \`SELECT karma_total, centroid_lat FROM community_observers_with_centroid LIMIT 0\` still resolves.

## Related

- Original fix: #547
- Regression: b6755ed
- Memory note: "CREATE OR REPLACE VIEW column-order trap — append-only rule"

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -4510,12 +4510,18 @@ SELECT
   expert_taxa, is_expert,
   observation_count, species_count, obs_count_7d, obs_count_30d,
   centroid_geog, last_observation_at, joined_at,
-  karma_total,
   -- Scalar lat/lng for clients that can't decode the geography WKB
   -- (the /community/map/ heatmap reads these directly). PostGIS
   -- geography → geometry cast is a zero-copy reinterpret for points.
   ST_Y(centroid_geog::geometry) AS centroid_lat,
-  ST_X(centroid_geog::geometry) AS centroid_lng
+  ST_X(centroid_geog::geometry) AS centroid_lng,
+  -- karma_total stays LAST per the append-only rule documented on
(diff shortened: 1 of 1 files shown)


ArtemioPadilla:
Superseded by #658 which landed the same fix. Closing.

PR #880, opened 2026-05-09
fix(social): optimistic revert on follow error (#877 followup)

ArtemioPadilla:
Fixup for #877 — follow button lacked the optimistic-revert path that the fave button already had.

**Bug:** on network/server error the follow state stayed as the optimistic value (e.g. `pending`) instead of reverting to the original (`none`).

**Fix:** capture `prevState` before the optimistic update, apply the transition to DOM immediately, revert to `prevState` in the catch block. Mirrors the exact pattern in the fave button handler.

Spotted by Nyx in cross-review. 🦉


ArtemioPadilla:
Merged into feat/709-like-follow-recents directly

PR #851, opened 2026-05-09
fix(observe): taxon filter chips give actionable feedback when cascade has no non-plant runner (#784)

ArtemioPadilla:
## Root cause

Taxon-hint chips (Insecto/Ave/Mamífero) were visually active (green) but appeared to have no effect. The cascade was correctly filtering PlantNet when the hint was `Animalia.*` — the problem was that:

1. `hideAllIdStates()` did not reset `gemmaOffer` / `gemmaProgress`. After a chip-triggered re-run, the Gemma download banner from the previous run could persist while the new run's state was hidden.

2. `describeAllFailed()` received the `runners` object **before** the hint filter was applied internally by `runParallelIdentify`. So `hadPlantNet: true` even when PlantNet was silently skipped. The resulting message was vague.

## Fix

- Added `gemmaOffer` and `gemmaProgress` to `hideAllIdStates()` reset sweep.
- Pre-apply `filterRunnersByHint()` in `triggerClientIdentify` before calling `runParallelIdentify`. Pass:
  - `hintFilteredRunners` — post-filter map for accurate `hadX` flags
  - `hintFilteredOutAll` — true when the hint removed the only available runner
- New branch in `describeAllFailed` for `hintFilteredOutAll + isAnimal`: targeted message explaining PlantNet is plant-only and directing user to Claude Haiku key or Gemma on-device.

The filter routing logic (`filterRunnersByHint`) was already correct — only user feedback was missing.

## Files
- `src/components/ObservationForm.astro`

## Test plan
- `npx tsc --noEmit`: no new errors
- `npm run test`: 1486 passed, 4 pre-existing gemma-vision failures
- Manual: classic form → load insect/bird photo → select 'Insecto' chip → should see targeted message if only PlantNet configured
- Manual: with Claude key → re-run should use Claude and return non-plant result

## v1.1.x follow-up
- #784 v1.1: add unit test for `describeAllFailed` hintFilteredOutAll branch in identify-cascade-client.test.ts

PR #901, opened 2026-05-09
fix(observe): unify filterRunnersByHint call + hintFilteredOutAll test

ArtemioPadilla:
Follow-up review fix for #851.


ArtemioPadilla:
Merged into fix/784-taxon-filter-noop directly

Issue #800, opened 2026-05-09
feat(notify): lunar event kairos trigger


ArtemioPadilla:
## Why (rationale)

Lunar phase is a known driver of biodiversity activity — full moons trigger anuran chorus peaks (some species call only during specific moon phases), influence sea-turtle nesting, predator-prey activity rhythms, nocturnal pollinator visits, and synchronised insect emergences. A simple opt-in push that fires on `luna llena` / `eclipse` / `luna nueva` for the user's timezone closes the lunar-event loop without any new external data — the `enrich-environment` Edge Function already computes lunar phase per observation. This was deferred from the v1 push so the golden-hour path could ship clean.

## Parent context

- Parent feature: PR #777 `feat(notify): golden hour contextual push prompts (kairos v1)` (shipped 2026-05-08 in v1.1.5)
- Module spec: `docs/specs/modules/34-kairos-prompts.md`
- Runbook: `docs/runbooks/kairos-prompts.md`
- Fogg principle: **Kairos** — Persuasive Technology (BJ Fogg, 2003), ch. 3

## Scope

- Add `kairos_subscriptions.kind = 'lunar_event'` opt-in toggle (default OFF) on `/profile/notifications`.
- Extend `kairos-fire` EF to fire on three lunar-event types: full moon, new moon, and lunar eclipse — when the event falls in the user's local timezone today.
- Reuse `enrich-environment`'s lunar phase computation (`supabase/functions/_shared/moon.ts` if it exists, otherwise extract from the EF). No new external API.
- Hard cap: **1 push/day total across all kinds** (existing `last_sent_at` row-lock).
- Body templates (EN/ES, fixed copy):
  - Full moon: `"Luna llena esta noche — coros de anfibios al máximo."` / `"Full moon tonight — peak amphibian chorus."`
  - New moon: `"Luna nueva — buena noche para insectos atraídos a luz."` / `"New moon tonight — good for light-trap insects."`
  - Lunar eclipse: `"Eclipse lunar visible esta noche."` / `"Lunar eclipse visible tonight."`

### Files to touch

- `supabase/functions/_shared/moon.ts` *(new or extracted)* — pure helpers `currentLunarPhase(date, lat, lng) → 'new'|'first_quarter'|'full'|'last_quarter'|'eclipse'|null` and `isLunarEventToday(now, tz) → 'full'|'new'|'eclipse'|null`. Should mirror SunCalc/Meeus algorithm already used for sunset.
[...]

PR #902, opened 2026-05-09
docs: fix lunar event window comment (#800 follow-up)

ArtemioPadilla:
Review fix: comment said ±12h but window is [midnight-12h, midnight+36h].


ArtemioPadilla:
Merged into feat/800-lunar-event-kairos directly

Issue #933, opened 2026-05-09
perf(auth): migrate all components to getCachedUser() — eliminate gotrue lock contention


ArtemioPadilla:
## Context

PR #930 introduced `getCachedUser()` in `supabase.ts` (30s in-memory cache, cleared on `visibilitychange: hidden`) and migrated `HomeWidgets.astro` to use it. The lock contention is reduced but not eliminated — 14+ other components still call `getSupabase().auth.getUser()` directly on mount.

## What to do

1. **Migrate all components** — replace `supabase.auth.getUser()` / `supabase.auth.getSession()` calls with `getCachedUser()` from `../lib/supabase`. Affected: ProfileView, ExploreRecentView, KarmaLeaderboardView, ObservationForm, MyObservationsView, and ~10 others.

2. **Add `SIGNED_OUT` invalidation** — add `onAuthStateChange` listener that sets `_userCache = null` on `SIGNED_OUT` event so sign-out reflects immediately.

## Acceptance
- `grep -rn 'auth\.getUser()' src/components/` returns 0 results
- Sign-out + reload shows signed-out state immediately
- gotrue lock warnings gone on profile page (Android Chrome 147)

## Related
- PR #930 (introduced getCachedUser, migrated HomeWidgets)
- Bug: 7× gotrue lock warnings on /perfil/ load

PR #993, opened 2026-05-11
fix(auth,onboard): callback spinner hang + onboarding tour for new signups

ArtemioPadilla:
## Summary

Two related bugs that combined to give brand-new signups the worst possible first session: stuck on the "Signing you in…" spinner forever, and even when they got through, the onboarding tour never appeared.

### Bug 1 — callback.astro hang

`redirectToHome()` was fired-and-forgotten in all 4 branches of `run()`. Because `run()` returned immediately, the outer `setTimeout(..., 8000)` was cleared by `run().finally(() => clearTimeout(timeout))` BEFORE the redirect actually completed. The 8 s safety net never fired.

Inside `redirectToHome()`, `supabase.auth.getUser()` was called directly — same gotrue navigator-lock contention surface that PRs #930 / #933 / #936 fixed elsewhere by introducing `getCachedUser()`. callback.astro was missed in that migration. When `getUser()` stalled, `window.location.replace(...)` was never reached.

**Fix:** `await redirectToHome()` in all 4 sites, switch to `getCachedUser()`, and wrap the `users` SELECT in a 3 s `withTimeout()` helper so the worst case is a redirect with `lang='en'`. The 8 s outer timeout now becomes a real safety net.

### Bug 2 — onboarding tour suppressed for new signups

`OnboardingTour.astro` `maybeShow()` checked `Object.keys(profile_privacy).length > 0` as a cross-device "user already configured privacy on another browser" signal. But the schema default for `users.profile_privacy` ships with all 19 keys pre-populated, so every brand-new user looked like an already-configured user → `markSeen()` → no tour, ever.

**Fix:** add `public.users.onboarding_completed_at timestamptz` (NULL for fresh users). Tour checks that instead. Written from `hide()` (both completed + dismissed paths) and from `persistPreset()`. `localStorage.rastrum.onboarding.seen` stays as the offline fast-path. No backfill needed — existing users who saw the tour already have the localStorage flag.

## Test plan

- [ ] db-validate CI passes (idempotent `ADD COLUMN IF NOT EXISTS` re-applies clean on the second pass)
- [ ] Manual: sign up with a fresh email → land on home in < 8 s → onboarding tour appears within 600 ms
- [ ] Manual: existing user with `localStorage.rastrum.onboarding.seen === 'v1'` → tour does NOT re-appear after this deploy
- [ ] Manual: throttle network in DevTools to simulate the original hang → callback shows the actionable error after 8 s instead of spinning forever
- [ ] Typecheck + vitest + build pass with no NEW failures (4 pre-existing TS errors in `sw-pmtiles-buffer.test.ts` and 6 pre-existing vitest failures unrelated to auth/onboarding remain the same — verified by diffing baseline)

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -2537,6 +2537,9 @@ ALTER TABLE public.users
   }'::jsonb,
   ADD COLUMN IF NOT EXISTS dismissed_privacy_intro_at timestamptz;
 
+ALTER TABLE public.users
+  ADD COLUMN IF NOT EXISTS onboarding_completed_at timestamptz;
+
 CREATE INDEX IF NOT EXISTS idx_users_profile_privacy
   ON public.users USING gin (profile_privacy jsonb_path_ops);
 
--- a/src/components/OnboardingTour.astro
+++ b/src/components/OnboardingTour.astro
@@ -271,6 +271,12 @@ const presetPrivate = onb.preset_private_observer ?? 'Private observer';
     try { localStorage.setItem(SEEN_KEY, SEEN_VAL); } catch { /* noop */ }
   }
 
+  // Guards: persistPreset() calls persistOnboardingDone() when the user
+  // picks a preset (they may close the modal before reaching the last step).
+  // hide() ALSO calls it when the modal actually closes. Both paths are safe
+  // because the UPDATE is idempotent — writing now() twice is a no-op effect.
+  // The paths are NOT mutually exclusive: preset → continue → last step → hide()
+  // will fire persistOnboardingDone() twice. This is intentional and harmless.
   async function persistOnboardingDone(): Promise<void> {
     let sb;
     try { sb = getSupabase(); } catch { return; }
--- a/src/components/OnboardingTour.astro
+++ b/src/components/OnboardingTour.astro
@@ -271,6 +271,16 @@ const presetPrivate = onb.preset_private_observer ?? 'Private observer';
     try { localStorage.setItem(SEEN_KEY, SEEN_VAL); } catch { /* noop */ }
   }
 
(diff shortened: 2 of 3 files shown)


Pamela-ruiz9:
## Review #993 — fix(auth,onboard): callback spinner hang + onboarding tour

LGTM. Dos bugs bien diagnosticados, dos fixes quirurgicos.

**Bug 1 — callback.astro hang**: el root cause esta bien identificado. fire-and-forget en las 4 ramas de run() + getCachedUser() vs getUser() era exactamente el mismo problema de gotrue lock contention. Los 4 sitios con await redirectToHome() son correctos.

withTimeout helper es el approach correcto para la query de users — 3 segundos es un fallback razonable (lang='en' es seguro como default cuando la query se cuelga). El helper hace resolve en ambos paths (timeout + error) sin rechazar, por lo que el outer try/catch no se complica. Bien.

**Bug 2 — onboarding tour suprimido**: el diagnostico es preciso. profile_privacy con 19 keys pre-poblados por schema default hacia que TODOS los usuarios nuevos parecieran configurados. El fix con onboarding_completed_at es la solucion correcta — campo nullable, NULL = nunca completado, timestamp = ya lo vio.

persistOnboardingDone() usa .catch() en ambos call-sites (hide + persistPreset) — correcto, ya que el tour debe cerrarse aunque la DB call falle.

El esquema add column IF NOT EXISTS es idempotente. No backfill necesario — los usuarios con localStorage tienen la fast-path, los sin localStorage en dispositivos nuevos simplemente veran el tour una vez y se guardara.

Una pregunta: el campo se escribe tanto en hide() (completed + dismissed paths) como en persistPreset(). Si un usuario hace preset pero NO cierra el modal normalmente, se escribe twice? No es un bug (idempotente), pero vale confirmar que ambos paths terminan siendo mutuamente excluyentes en la practica.

Issue #914, opened 2026-05-09
chat: location entity kind (waits for locations-first-class table)


ArtemioPadilla:
v1.1 entity registry has 6 kinds (observation/species/project/camera_station/observer/self_profile). `location` was deferred because `public.locations` doesn't exist yet (locations-first-class spec at `docs/superpowers/specs/2026-05-03-locations-first-class-design.md` not implemented).

## Scope (after locations table lands)
- Add `src/lib/chat-entities/location.ts` EntitySpec
- Add `location` case to `chat_entity_card` dispatcher in supabase-schema.sql
- Add `chat_location_card(p_id uuid)` SECURITY INVOKER function
- Add `location` tab to `ChatEntityPicker`
- Add `AskRastrumButton` to location detail surface

## Blocked by
locations-first-class spec implementation.

Module: 20-chat.

Issue #866, opened 2026-05-09
feat(retention): streak freeze / skip-day mechanic — anti-rage-quit


ArtemioPadilla:
## Context

Strategic gap. Today `users.streak_current` resets to 0 on any miss (the `recompute-streaks` cron walks `karma_events` per user and resets when the day-gap >1). For a community where users' field outings are weather-, work- and school-dependent, a single bad day rage-quits the streak — which Duolingo explicitly identified as one of their highest-impact retention regressions before introducing streak freezes.

This issue introduces a small, transparent freeze mechanic with hard caps to prevent abuse.

## Why

- Duolingo's streak freeze is one of the highest-impact retention features in mobile apps (reportedly responsible for ~13% lift in 7-day retention on the streak cohort). Rastrum has the same loss-aversion exposure with a weaker safety net (zero).
- Rage-quit risk is real for casual observers — losing a 30-day streak after one rainy weekend is exactly the friction we want to avoid.
- Fogg's *Persuasive Technology*: never punish the user for a contextual miss when a low-cost forgiveness mechanic is available — the punishment cost > the motivation gain.

### Server-side

- New column `users.streak_freezes_available smallint NOT NULL DEFAULT 0`. Hard cap at **2** to prevent abuse (cron clamps to `LEAST(streak_freezes_available + 1, 2)`).
- New column `users.streak_freezes_used integer NOT NULL DEFAULT 0` (lifetime count, transparency).
- Extend the existing `recompute-streaks` cron handler:
  - When a user crosses a 7-day-streak boundary (i.e. `streak_current % 7 == 0` AND `streak_current > 0` from previous run), grant +1 freeze (clamped at 2).
  - When detecting a missed day:
    1. If `streak_freezes_available > 0`: decrement, increment `streak_freezes_used`, **preserve** `streak_current`, insert a `karma_events` row with `delta = 0` and a new `outcome` value `streak_freeze_consumed`. Do NOT reset.
    2. If `streak_freezes_available == 0`: existing reset behavior.
- Schema invariants: `LANGUAGE plpgsql SECURITY DEFINER SET search_path = public, extensions, pg_temp`, `REVOKE EXECUTE FROM PUBLIC` on any new function, `GRANT EXECUTE TO service_role`. Idempotent re-apply.
- Add `outcome = 'streak_freeze_consumed'` to the existing `karma_events.outcome` enum or CHECK constraint (depending on current schema shape).
[...]

Issue #803, opened 2026-05-09
feat(suggest): probable_taxa_cache layer


ArtemioPadilla:
## Why (rationale)

The contextual-species chips on `/observe` (PR #774) compute their suggestions **live** every time the user opens the form — `probable_taxa_at(lat, lng, month, limit)` runs an indexed `ST_DWithin(50 km)` join over `observations` and groups by taxon. At today's scale (~hundred-thousand obs) this is well under 300 ms on the GIST index. As the dataset grows past 1 M observations, the live-grouping cost grows roughly linearly and we'll start hitting 800 ms - 1.5 s on first paint. The contextual chips are **above the fold** in `/observe`, so any latency here directly compresses the Fogg "Reduction" win we just shipped. The cache layer was sketched in the PR but deferred because it's premature optimisation today; the runbook spells out the trigger.

## Parent context

- Parent feature: PR #774 `feat(suggest): contextual species suggestions by location + season` (shipped 2026-05-08 in v1.1.5)
- Module: M01 (Photo ID) + M02 (Observation Form)
- Runbook: `docs/runbooks/contextual-suggestions.md` → "Cache layer (deferred)" section
- Fogg principle: **Reduction** (suggesting a quick taxon match) — Persuasive Technology (BJ Fogg, 2003), ch. 3

## Scope

- New `probable_taxa_cache (geohash5, month, taxon_id, rank, n_obs, refreshed_at)` materialised view OR plain table populated by a nightly cron `refresh-probable-taxa-cache`.
- Update `probable_taxa_at()` RPC to read from the cache first (geohash5 + month lookup), falling back to the live aggregate only on cache miss (e.g. user in remote region with no cached row).
- Cron schedule: nightly at 09:00 UTC (after `recompute-user-stats` finishes at 08:00) so cron load is staggered. Refresh full table; keep it small (32 MX states × 12 months × top-50 taxa × ~1k geohash5 cells = ~20M rows max).
- Cache miss telemetry: emit a structured log `function_errors` event with severity=info if a user query falls through to live — operator can monitor cache coverage.
- Triggers documented in the runbook: dataset > 1 M obs, OR per-pan map hot path latency complaints.

### Files to touch

[...]

Issue #550, opened 2026-05-04
feat(karma): IUCN + NOM-059 conservation status ETL pipeline


ArtemioPadilla:
## Context

`taxa.iucn_category` and `taxa.nom059_status` columns already exist in `docs/specs/infra/supabase-schema.sql:177` (and a documentation marker at `docs/specs/infra/supabase-schema.sql:6772-6784`), and `src/lib/karma-conservation.ts` defines the multiplier lookup table the karma path is supposed to consume. What's missing is the **data**: there is no pipeline that populates these columns from authoritative sources, so they are NULL for almost every taxon and the conservation bonus stays inert.

This unblocks wiring the multipliers into `award_karma()` and restoring the conservation suffix in `microcopyForVote()` (PR #544 stripped that suffix because it predicted a bonus the SQL wasn't actually applying).

## Acceptance criteria

- [ ] One-time backfill script (Edge Function or Node script under `scripts/`) hydrates `taxa.iucn_category` from the GBIF Species API for every taxon with `gbif_taxon_key IS NOT NULL`
- [ ] Backfill also hydrates `taxa.nom059_status` from CONABIO's published NOM-059 species list (Mexico-only — most taxa stay NULL)
- [ ] Nightly refresh as a cron Edge Function (e.g. `supabase/functions/refresh-conservation-status/`) re-pulls a delta and updates only changed rows; idempotent
- [ ] Operator runbook at `docs/runbooks/conservation-status-etl.md` documents source URLs, rate limits, and how to fire the backfill manually
- [ ] CI smoke check (or runbook step) verifies coverage: at least N% of synced-observation taxa have a non-NULL conservation status
- [ ] No-op when GBIF/CONABIO endpoints are unreachable — log + continue, never wipe existing data

## Files / code pointers

- `docs/specs/infra/supabase-schema.sql:177` — column definitions on `taxa`
- `docs/specs/infra/supabase-schema.sql:6772-6784` — existing documentation marker
- `src/lib/karma-conservation.ts` — multiplier lookup table consumers will read
- `supabase/functions/award-badges/index.ts:11` — already references `nom059_any` rule, will benefit from populated data
- `supabase/functions/_shared/cron-auth.ts` — cron secret pattern for the new Edge Function

## Dependencies

Blocks the conservation-multiplier wiring (a separate issue will be filed and marked `Depends on` this one).

## Out of scope

- Wiring the multipliers into `award_karma()` SQL (separate follow-up issue)
[...]

PR #994, opened 2026-05-11
fix(ci): unblock db-validate (storage.foldername stub + places column alias)

ArtemioPadilla:
## Summary

\`db-validate\` has been broken on every recent PR for two compounding reasons. This PR resolves both:

### 1. \`storage.foldername(text)\` missing in CI Postgres

The ephemeral container lacks Supabase's built-in. \`media_authenticated_list\` policy referenced it at \`supabase-schema.sql:1326\` and blew up the schema apply.

**Fix:** add a minimal SQL equivalent (\`string_to_array(name, '/')\`) inside the existing "Stub Supabase auth schema and helpers" step.

### 2. \`chat_find_location\` references a non-existent column

After #1 unblocks the apply past line 1326, validate then fails at line 10684:

\`\`\`
ERROR: column "observation_count" does not exist
\`\`\`

Root cause (not introduced by this PR — discovered by it): \`docs/specs/infra/supabase-schema.sql\` has **two** \`CREATE TABLE IF NOT EXISTS public.places\` blocks:

| Line | Origin | Columns |
|---|---|---|
| ~7285 (wins via \`IF NOT EXISTS\`) | older WDPA / protected-areas table | \`obs_count\`, \`species_count\`, \`observer_count\` |
| ~10650 (silent no-op) | #914 chat-locations | \`observation_count\`, \`observer_count\`, \`top_taxa\` |

\`chat_find_location\` (added in #914) selects \`observation_count\`, but the table that actually exists in CI (and in production) spells the column \`obs_count\`.

**Fix:** alias \`obs_count AS observation_count\` in the SELECT and \`ORDER BY obs_count\` in the order clause. The function's \`RETURNS TABLE (..., observation_count int)\` signature is preserved — clients keep seeing \`observation_count\`. The deeper question of which \`places\` design ultimately wins is left to a follow-up PR with @ArtemioPadilla (the #914 author).

## Why these are bundled

Both fixes are required for \`db-validate\` to ever go green again. Splitting them would leave the gate broken until both merge anyway. Reviewable in one pass: 9 inserted lines in db-validate.yml + 7 inserted / 2 deleted in supabase-schema.sql, both with explanatory comments.

## Test plan

- [ ] \`db-validate\` workflow on this PR reaches the second-pass idempotency check, the sentinel-verify step, the RLS suite, and the schema-security lints — all green
- [ ] After merge, the next PR that touches \`supabase-schema.sql\` actually gets gated by validate again (it has been silently broken-by-design)
- [ ] No production data change — \`places\` table contents unaffected, only the function's SELECT changes
- [ ] \`chat_find_location('mexico')\` still returns rows with a column named \`observation_count\` in the response

### DIFF
--- a/.github/workflows/db-validate.yml
+++ b/.github/workflows/db-validate.yml
@@ -134,6 +134,15 @@ jobs:
           GRANT ALL ON storage.buckets, storage.objects TO authenticated, service_role;
           GRANT SELECT ON storage.buckets, storage.objects TO anon;
 
+          -- storage.foldername(text) — Supabase's real impl splits a path by '/' and
+          -- returns the array of folder segments minus the final filename. For the
+          -- validate context we just need a function that lets the policy compile
+          -- and evaluate correctly; the simplest equivalent splits on '/' and returns
+          -- all segments, which the only consuming policy ([1] != '.keep') reads
+          -- positionally from. See docs/specs/infra/supabase-schema.sql media_authenticated_list.
+          CREATE OR REPLACE FUNCTION storage.foldername(name text) RETURNS text[]
+          LANGUAGE sql IMMUTABLE AS $$ SELECT string_to_array($1, '/') $$;
+
           -- Cron stubs MUST be created before supabase-schema.sql is applied,
           -- because rastrum's schema now defines SQL functions whose body
           -- references cron.job and cron.job_run_details (see
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -12827,13 +12827,14 @@ CREATE UNIQUE INDEX IF NOT EXISTS mv_user_obs_counts_idx
 -- Avoids a full-table scan on the explore / trending-species page.
 CREATE MATERIALIZED VIEW IF NOT EXISTS public.mv_recent_species AS
   SELECT
-    taxon_id,
+    primary_taxon_id AS taxon_id,
     COUNT(*) AS recent_count,
     MAX(observed_at) AS last_seen
   FROM public.observations
   WHERE sync_status = 'synced'
     AND observed_at > now() - interval '30 days'
-  GROUP BY taxon_id
+    AND primary_taxon_id IS NOT NULL
+  GROUP BY primary_taxon_id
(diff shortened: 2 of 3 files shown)


ArtemioPadilla:
## Close-out — split into 9 per-author follow-up PRs

This PR accumulated 10 commits fixing bugs from 6 different recently-merged PRs while peeling layers of compounding CI breakage. That was a scope mistake on my part — each layer needs the original author's review for proper accountability.

**Split into individual PRs**, each one concern, each linked to the original bug-introducing PR:

| # | Concern | Origin |
|---|---|---|
| #1005 | CI: \`storage.foldername\` stub in db-validate Postgres env | (foundation — merge first) |
| #1006 | \`chat_find_location\` uses \`obs_count\` (places column rename) | #914 |
| #1007 | Remove dead \`idx_activity_target_unread\` (column never existed) | — |
| #1008 | Cron #550 \`$$\`-quote nesting → \`$do$\`/\`$cron$\` tagged | #550 |
| #1009 | Close \`institutions\` + \`weekly_validator_rewards\` CREATE TABLE | #997 |
| #1010 | \`probable_taxa_cache\` policy missing DROP IF EXISTS | #803 |
| #1011 | trails/pits policies missing DROP IF EXISTS (×4) | #995 |
| #1012 | rls.sql #866 freeze test missing auth.users seed | — |
| #1013 | \`mv_recent_species\` SELECT uses \`primary_taxon_id\` (column rename) | #1002 |

**Merge order:** #1005 first (foundation). The other 8 each fail validate in isolation but pass once #1005 is in. Validate becomes a real gate again.

## Best-practice notes for next time

What I learned doing this:
1. **Don't bundle bugs from multiple authors in one PR.** Each one needs its own blame trail and reviewer.
2. **Don't peel cascading layers in a force-pushed branch.** The CI cycles are long; better to fix the foundation, merge, and re-run.
3. **Don't add band-aid skips** (e.g., test.skip) to make tests pass — file an issue and fail loud.
4. **A required CI check that's path-filtered out** is a branch-rule misconfiguration; needs fixing structurally, not via merge-with-override.

Closing this with no commits lost — all 10 commits live across the 9 sibling PRs.

PR #996, opened 2026-05-11
fix(test): tighten Mock postMessage type in sw-pmtiles-buffer.test

ArtemioPadilla:
## Summary

`tests/unit/sw-pmtiles-buffer.test.ts` declared its mocked `postMessage` as `ReturnType<typeof vi.fn>`, which resolves to `Mock<Procedure | Constructable>`. TypeScript correctly refuses to consider that assignable to the function-under-test's expected `(...a: unknown[]) => void` parameter, producing 4 TS2345 errors at lines 86 / 95 / 96 / 105 — and breaking the `verify` workflow's typecheck on every PR.

Fix by passing the consumer signature as the type parameter to `vi.fn<...>()`. The Mock now narrows to `Mock<PostMessageFn>`, which is assignable to the consumer's parameter. `.mock.calls` access stays intact (Mock keeps its tracking properties either way).

No runtime change. No production-code change. Pure types-only fix.

## Test plan

- [ ] `npx tsc --noEmit` no longer reports errors in `tests/unit/sw-pmtiles-buffer.test.ts`
- [ ] `npx vitest run tests/unit/sw-pmtiles-buffer.test.ts` passes all sub-tests
- [ ] verify workflow goes green

### DIFF
--- a/src/lib/chat-tools.test.ts
+++ b/src/lib/chat-tools.test.ts
@@ -10,10 +10,11 @@ beforeEach(() => {
 });
 
 describe('chat-tools', () => {
-  it('exposes 5 tools', () => {
+  it('exposes 6 tools', () => {
     const names = listTools().map(t => t.name).sort();
     expect(names).toEqual([
       'find_camera_stations',
+      'find_location',
       'find_observations',
       'find_observers',
       'find_projects',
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
Closing in favor of #999, which bundles this PR's two fixes with #999's build fix. #999 and #996 were mutually deadlocked — `verify` on #999 fails on this PR's typecheck issues, and `test` on this PR fails on #999's build issue. Neither could go green alone. Consolidating per the user's 'best UX' directive.

PR #1005, opened 2026-05-11
fix(ci): stub storage.foldername() in db-validate Postgres env

ArtemioPadilla:
**Foundation fix — merge this first.** All other "fix(schema)" PRs extracted from #994 close-out depend on validate being able to advance past line 1326.

## Summary

\`db-validate\` has been failing on every PR for weeks at \`supabase-schema.sql:1326\` because the ephemeral Postgres container lacks Supabase's \`storage.foldername(text)\` — referenced by the \`media_authenticated_list\` policy. Once that line errors, validate exits before doing any real validation.

Add a minimal SQL equivalent (\`string_to_array(name, '/')\`) inside the existing "Stub Supabase auth schema and helpers" step in \`.github/workflows/db-validate.yml\`. The only consumer reads index \`[1]\` to gate \`.keep\` sentinel files; split-on-'/' is sufficient.

## Why this matters

The validate gate has been broken-by-design since this stub was missing. Multiple PRs (#914, #550, #803, #995, #997, #1002) merged with schema bugs that should have been caught here. Restoring the gate unblocks 8 sibling fix PRs that revert those bugs.

## Test plan

- [ ] \`db-validate\` advances past line 1326 on this PR
- [ ] After merge, sibling cleanup PRs surface remaining schema bugs one by one

Extracted from #994 (close-out — 9 separate per-author PRs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1006, opened 2026-05-11
fix(schema): chat_find_location uses places.obs_count (not observation_count)

ArtemioPadilla:
## Summary

\`chat_find_location\` (added in #914) does \`SELECT … observation_count FROM public.places\`, but the existing \`places\` table (~line 7285, WDPA-derived, predates #914) spells the column \`obs_count\`. #914's own \`CREATE TABLE IF NOT EXISTS places\` (~line 10650) is a silent no-op.

**Fix:** alias \`obs_count AS observation_count\` in the SELECT and \`ORDER BY obs_count\`. Preserves the function's \`RETURNS TABLE (..., observation_count int)\` contract — clients keep seeing \`observation_count\`.

## Out of scope

The deeper conflict — two distinct \`places\` table designs in the schema — needs the #914 author's call. This PR is the minimum to unblock validate.

## Test plan

- [ ] db-validate advances past line 10684 (requires #1005 merged first)
- [ ] \`chat_find_location('mexico')\` still returns rows with \`observation_count\` column

Extracted from #994 close-out. Original bug: #914.


ArtemioPadilla:

PR #1008, opened 2026-05-11
fix(schema): nested dollar-quoting in refresh-conservation-status cron block

ArtemioPadilla:
## Summary

PR #550 added the \`refresh-conservation-status\` cron schedule inside a \`DO \$\$ … \$\$\` block. Both the outer DO body AND the inner \`cron.schedule\` command literal use the default \`\$\$\` delimiter — so the SQL parser treats the inner \`\$\$SELECT net.http_post(...)\` opening as the close of the outer DO body, yielding "syntax error at or near SELECT" against the matching outer \`\$\$\` later in the file.

**Fix:** tag the outer block as \`\$do\$ … \$do\$\` and the inner command as \`\$cron\$ … \$cron\$\`. Same effective semantics; unambiguous parse.

## Test plan

- [ ] db-validate advances past line 12109 (requires #1005 merged first)
- [ ] \`SELECT * FROM cron.job WHERE jobname = 'refresh-conservation-status'\` returns the same row contents

Extracted from #994 close-out. Original bug: #550.

### DIFF
--- a/docs/specs/infra/supabase-schema.sql
+++ b/docs/specs/infra/supabase-schema.sql
@@ -12085,21 +12085,21 @@ ALTER TABLE public.taxa
 -- Monthly cron: triggers the refresh-conservation-status Edge Function.
 -- Runs on the 1st of each month at 03:00 UTC.
 -- Requires pg_cron + pg_net extensions (already enabled in Rastrum).
-DO $$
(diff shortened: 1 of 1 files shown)


ArtemioPadilla:

PR #1010, opened 2026-05-11
fix(schema): probable_taxa_cache CREATE POLICY missing DROP POLICY IF EXISTS

ArtemioPadilla:
## Summary

Per CLAUDE.md "Idempotent everything", every \`CREATE POLICY\` must be paired with a preceding \`DROP POLICY IF EXISTS\`. \`probable_taxa_cache\`'s policy (~line 11823) ships from #803 without the DROP guard — validate's second-pass idempotency check fails with "policy already exists".

**Fix:** one-line \`DROP POLICY IF EXISTS\` before the \`CREATE POLICY\`.

## Test plan

- [ ] db-validate's second pass (idempotency) advances past line 11823 (requires #1005 merged first)

Extracted from #994 close-out. Original bug: #803.


ArtemioPadilla:

PR #974, opened 2026-05-10
feat: Kairos day-3, provisional first-ID, onboarding demo, Discover feed, Play Store CI (#897 #898 #896 #865 #871)

ArtemioPadilla:
- **#897** Kairos push at day 3 for users with 0 observations\n- **#898** Provisional ID (0.3 threshold) for new users with 'awaiting confirmation' label\n- **#896** Guided first-observation demo step in onboarding tour\n- **#865** Personalized Discover feed (follows + nearby + trending), EN+ES pages, nav updated\n- **#871** Tauri AAB sign + Play Store internal-track upload CI on tag push\n\nCloses #897, #898, #896, #865, #871

PR #1020, opened 2026-05-11
fix(e2e): onboarding tour step count 6 → 7 (after first-observation-demo step)

ArtemioPadilla:
## Summary

OnboardingTour grew from 6 to 7 steps when #974 added the \`first_observation_demo\` step. Two e2e specs were not updated and have been failing since #1015 restored the e2e workflow:

- \`tests/e2e/onboarding.spec.ts\` — 3 tests asserted \`/Step N of 6/\`
- \`tests/e2e/journey-observer-first-obs.spec.ts\` — loop iterated 6 times

Verified against \`src/components/OnboardingTour.astro\`: \`data-step-count="7"\` and 7 entries in the stepsData JSON literal.

Also tightened the \`journey-observer\` assertion from \`toContainText(\\\`\${i + 1}\\\`)\` (loose substring match, would pass coincidentally on digit overlap) to \`toContainText(\\\`\${i + 1} of 7\\\`)\` (verbatim format) — so a future drift to 8 steps fails loudly.

## Closes / supersedes

Closes #999 (which had this fix among other commits, most of which became redundant after #1018, #1019, #1004 merged equivalent fixes elsewhere; this is the only remaining real delta).

## Test plan

- [ ] CI \`test\` workflow (Playwright e2e) green
- [ ] All 5 specs in \`tests/e2e/onboarding.spec.ts\` pass
- [ ] J2 journey \"onboarding tour opens and can be completed\" passes
- [ ] No regression in passing sibling tests

### DIFF
--- a/tests/e2e/journey-observer-first-obs.spec.ts
+++ b/tests/e2e/journey-observer-first-obs.spec.ts
@@ -14,9 +14,9 @@ test.describe('J2: Observer first observation journey', () => {
     const dialog = page.locator('#onboarding-tour');
     await expect(dialog).toBeVisible();
 
-    // Walk through all 6 steps
-    for (let i = 0; i < 6; i++) {
-      await expect(page.locator('#onb-step-label')).toContainText(`${i + 1}`);
+    // Walk through all 7 steps
+    for (let i = 0; i < 7; i++) {
+      await expect(page.locator('#onb-step-label')).toContainText(`${i + 1} of 7`);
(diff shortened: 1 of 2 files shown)


ArtemioPadilla:
Superseded — #999 already merged with the same fix. main has both onboarding.spec.ts step count 6→7 and journey-observer-first-obs.spec.ts loop 6→7. Closing as redundant.

Issue #1031, opened 2026-05-11
epic(qa): programmatic confidence playbook — minimize manual testing across schema, frontend wiring, security, UX, perf


ArtemioPadilla:
> **Why this exists.** Manual testing every feature is the single biggest drag on
> this repo. The intentionally-minimal e2e suite (per CLAUDE.md) was the right
> call at PR #50; at PR #1029 the gap shows: 6 weeks of accumulated schema bugs
> hid behind a broken validate gate (#1015 close-out), 3 unit tests fail locally
> in CST while passing in UTC CI, and several "shipped clean" PRs broke
> integration paths nothing tests. This umbrella issue is the plan to close the
> programmatic-confidence gap — feature actually works end-to-end, integration
> wired, security holds, UX/perf don't regress — without per-PR human QA.

## Reality check

Some things can't be automated. The plan covers ~80% of what we currently test
by hand. The remaining ~20% (does the copy feel right? is the abstraction
correct? is onboarding converting?) always needs a human or PostHog. Don't
chase 100%; chase the 80% that's mechanical.

## Diagnosis (verified 2026-05-12)

| Layer | Coverage today | Real protection |
|---|---|---|
| Schema syntax + idempotency | ✅ `db-validate.yml` (restored #1015) | Strong |
| Schema security invariants | ✅ `infra/lint-schema-security.sql` (#1029) | Good |
| TypeScript correctness | ✅ `tsc --noEmit` | Strong |
| Astro `define:vars` + dynamic import trap | ✅ `scripts/check-define-vars-imports.sh` | Strong |
| Search-index drift | ✅ wired in `ci.yml` | Strong |
| Unit logic (Vitest) | 🟡 2356 tests / 205 files; coverage unmeasured | OK but unknown |
| RLS / privilege escalation | 🟡 55 assertions in `tests/sql/rls.sql` | Partial |
| Edge Function shared helpers | 🟡 5 unit tests (`_shared/*`) | Partial |
| Edge Function HTTP handlers | ❌ none | **Zero** |
| Vision provider live probe | ✅ `vision-providers-smoke.yml` (nightly, 7 providers) | Strong |
| E2E journeys | 🟡 11 journey specs + 24 other e2e specs (35 total) | Partial — see gaps below |
| TZ / locale determinism of unit tests | ❌ no enforcement; 3 tests pass in UTC CI, fail in CST | **Zero** |
| Visual regression | 🟡 PostHog VR runs, not a required check | Partial |
[...]

PR #1049, opened 2026-05-14
fix(date): TZ-determinism — UTC accessors for ISO weeks + phenology + test pin (#1031 Tier 1c)

ArtemioPadilla:
## Summary

Tier 1c of #1031 — TZ/locale determinism. Fixes 3 unit tests that fail on CST laptops but pass in UTC CI, plus the underlying production bug they expose.

- **Production fix:** `cohortWeek()` in `src/lib/posthog-events.ts` and the phenology chart in `SpeciesProfileView.astro` both leaked local-TZ accessors when computing what should be timezone-invariant ISO weeks / month histograms. Users in non-UTC timezones saw cohort weeks off-by-one at year boundaries and monthly observation counts off-by-one for observations near month boundaries.
- **Test config:** `vitest.config.ts` now pins `process.env.TZ ??= 'UTC'` + `test.env: { TZ: 'UTC' }` (belt-and-braces — `test.env` injects via `import.meta.env`, the `process.env` line catches Node Date construction in worker init). `playwright.config.ts` pins `timezoneId: 'UTC'` + `locale: 'en-US'`.
- **Test helpers:** 4 test files (`species-explorer-p3`, `expert-lottery`, `observador-del-mes`, `kairos-migration-window`) had their date helpers switched to UTC accessors. All mirror production code that's already UTC-safe (or now is, post-fix).
- **CLAUDE.md:** Added a "Pitfalls discovered 2026-05-13" row documenting the pattern + what's intentionally LEFT local (`todayKey`, `pickSeasonalTheme`, `isoToLocalDatetimeInput` — all user-facing surfaces where the user's local TZ is the right answer).

## What was deliberately NOT changed

- `src/lib/surprises.ts` `todayKey()` — daily field-trip cap matches user's local day.
- `src/lib/seasonal-theme.ts` `pickSeasonalTheme()` — user's local month is the intent.
- `src/lib/manage-panel.ts` `isoToLocalDatetimeInput()` — explicit UTC→local for `<input type="datetime-local">`.
- `src/lib/suggestions.ts:20` — user-facing seasonal "what's common this month".
- `src/components/home/HomeObservadorDelMes.astro` — user-facing "this month's featured observer". (Worth a follow-up discussion whether the monthly-rotation key should be operator-controlled UTC, but out of scope here.)
- `cli/src/batch-import.test.ts` — EXIF dates are local-by-spec.

The vitest pin is the safety net for future regressions: any new code that introduces this class of bug fails the same way locally as it would in production for non-UTC users.

## Test plan

- [x] `TZ=America/Chicago npm test` — 2356/2356 pass on this branch (was 3 failed / 2353 passed on main).
- [x] `TZ=UTC npm test` — 2356/2356 pass.
- [x] `npx tsc --noEmit` — clean.
- [ ] CI green (typecheck, vitest, e2e, build).
- [ ] After merge: spot-check the phenology chart on `/share/obs/...` for a species with March/June observations from a non-UTC laptop.

Refs #1031 (Tier 1c).


ArtemioPadilla:
**Update (review follow-up).** Reviewer pointed out that the `expert-lottery` test now asserts UTC behavior but I hadn't fixed the production EF it mirrors. Verified + fixed in `8641343`:

- **`supabase/functions/weekly-expert-lottery/index.ts` `getIsoWeek()`** — same class of bug (`setDate`/`getDate`/`getDay`/`getFullYear` instead of UTC variants). Works in production today because Supabase Edge runs UTC, but the test would have misrepresented prod behavior if anyone ever ran the EF locally on a non-UTC machine.

**Bonus findings — three lower-severity TZ leaks in other EFs** I noticed while sweeping. Not blocking this PR (production Edge=UTC hides them), but worth a separate fix:

- `supabase/functions/_shared/sponsorship.ts:101,132` — `new Date(new Date().getFullYear(), new Date().getMonth(), 1).toISOString()` for `monthStart` in billing queries. Off by ~6h on local-dev CST machines.
- `supabase/functions/sponsorships/index.ts:546` — same pattern.
- `supabase/functions/generate-wrapped/index.ts:51` — `new Date().getFullYear()` for "year must not be in future" check; edge case at the Dec/Jan boundary in non-UTC TZ.

Happy to roll these into this PR if reviewer prefers; otherwise will file a follow-up issue after merge.

PR #1068, opened 2026-05-16
fix(cron): weekly-digest + refresh-conservation-status use Vault, not unset GUCs

ArtemioPadilla:
## Problem

Two pg_cron jobs defined in `supabase-schema.sql` built their `net.http_post` call from `current_setting('app.supabase_url')` / `current_setting('app.cron_secret')`. **Neither GUC was ever set** on the database, so every run failed in Postgres with:

```
ERROR: unrecognized configuration parameter "app.supabase_url"
```

The Edge Function was never even called.

- `weekly-digest` — hourly (`0 * * * *`)
- `refresh-conservation-status` — monthly (`0 3 1 * *`)

`weekly-digest` additionally sent the secret as `Authorization: Bearer …`, but the EF gates on the **`X-Cron-Secret`** header (`requireCronSecret`) — so it would 403 even if the GUC existed.

Discovered while resolving the cron→EF pipeline incident (#1066 / CRON_SECRET bootstrap / #1067). Independent of those — this is a Postgres-side error that fires before any HTTP call.

## Fix

Rewrite both to the **canonical cron→EF shape** used by every working job in `cron-schedules.sql`:

- Hardcoded function URL (`https://reppvlqejgoqvitturxp.supabase.co/functions/v1/<name>`)
- `X-Cron-Secret` header read from `vault.decrypted_secrets WHERE name = 'cron_secret'` at query time

No external GUC state that can be forgotten. Idempotent: `cron.schedule` upserts by name; the conservation block keeps its existing `pg_cron`-guarded unschedule/schedule.

## Test plan

- [ ] `db-validate` gate (ephemeral PG, double-apply) green
- [ ] After merge → `db-apply` → next ticks: `weekly-digest` (hourly) returns 200 in `net._http_response`; `refresh-conservation-status` verified at next monthly run or via manual fire

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1066, opened 2026-05-16
fix(deploy): cron-only EFs must skip Edge-layer JWT verification

ArtemioPadilla:
## Incident

Every `pg_cron` → Edge Function call has been returning **401 at the Supabase gateway** (never reaching function code), silently, for days. Discovered while investigating why the GBIF enrich cron was red.

`cron.job_run_details` reports these jobs as `succeeded` — misleading, because `net.http_post` only **enqueues** (pg_net is async). The real HTTP status is in `net._http_response`, which showed **401/500 across the board, zero 200s**.

Affected jobs: `streaks-nightly`, `badges-nightly`, `recompute-user-stats-nightly`, `streak-push-nightly`, `plantnet-quota-daily`, `kairos-fire-15min`, `retry-unidentified`.

## Root cause

Cron-only EFs self-gate on `X-Cron-Secret` via `_shared/cron-auth.ts` (`requireCronSecret`). pg_cron calls them with **no `Authorization` bearer**, so they must be deployed `--no-verify-jwt` or the gateway 401s *before* `requireCronSecret` ever runs. Two config drifts caused regressions:

- `supabase/config.toml` had `verify_jwt = true` for `recompute-streaks` and `award-badges`, contradicting the deploy workflow's `CRON_ONLY` classification.
- `deploy-functions.yml` `CRON_ONLY` omitted `kairos-fire`, `recompute-taxa-cache`, `weekly-digest` (all use `requireCronSecret`).

Since `deploy-functions.yml` only redeploys *changed* functions on push, EFs deployed before being classified kept their stale `verify_jwt=true` gateway setting in prod.

## Changes

- `config.toml`: `recompute-streaks` + `award-badges` → `verify_jwt = false`; added explicit `false` blocks for `kairos-fire` / `recompute-taxa-cache` / `weekly-digest`; documented the both-sites invariant. `enrich-environment` deliberately left `true` (invoked from sync pipeline with a real JWT, not cron).
- `deploy-functions.yml`: added the three missing EFs to `CRON_ONLY`.

Editing `deploy-functions.yml` trips `WORKFLOW_TOUCHED` → **merge auto-redeploys all functions** with the corrected classification.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run test` — 2358 passed
- [ ] After merge: confirm deploy-functions.yml run is green
- [ ] After deploy: re-query `net._http_response` — expect 200s (or 403 → separate `CRON_SECRET`/Vault follow-up)

[...]

Issue #868, opened 2026-05-09
feat(notify): weekly email digest for inactive users


ArtemioPadilla:
## Context

Strategic gap. The kairos push infrastructure (#724/#798/#799/#800) reaches users who have notifications enabled and are still installed. It does NOT reach users who:
- Disabled push notifications.
- Cleared the PWA install / are using a non-PWA browser.
- Are early-stage churned (no observation in 7+ days).

Email is the second-funnel for re-engaging these users. Note: #559 already exists as a **karma-only** weekly digest (parking-lot, narrower scope — only summarizes the user's own karma earned that week). This issue is the broader **re-engagement** digest aimed at inactive users with personalized recommendations from the social graph.

## Why

- Push reach drops to ~30-50% on iOS Safari and is ~0% for users who never enabled it. Email is the only channel that reaches everyone with an account.
- Duolingo, Strava, and iNaturalist all run weekly digests as the primary inactive-user recovery mechanic. Reported lifts in the 5-15% range on 30-day return rate.
- Fogg's *Persuasive Technology*: contextual triggers (push) for active users, scheduled triggers (email) for inactive ones. Different cohorts need different prompts.

### Server-side

- New cron-only Edge Function `supabase/functions/weekly-digest/index.ts`.
  - Deployed `--no-verify-jwt` and gated by `requireCronSecret` (mirror `recompute-streaks`).
  - Runs **Sunday at the user's local 14:00** — schema invariant: use `users.timezone` (already added per CLAUDE.md PR14). Compute the recipient set per UTC hour and dispatch only the users whose local time matches the current cron tick.
- Targets: users where `last_observation_at < now() - interval '7 days'` AND `email_notifications_enabled = true`.
- New column `users.email_notifications_enabled boolean NOT NULL DEFAULT true` (or read from the granular `notification_prefs` JSONB defined in #20 — coordinate with that issue; whichever ships first defines the contract).
- Content per user (rendered in plain text + HTML):
  - Their rank delta vs last week (read `community_observers` snapshot from `recompute-user-stats` cron output).
  - Top 3 observations from people they follow (`SELECT … FROM observations WHERE observer_user_id IN (SELECT following_id FROM follows WHERE follower_id = $user) ORDER BY likes_count DESC, created_at DESC LIMIT 3`).
  - 1 nearby species they haven't observed yet (reuse `falta-dex` RPC if available — see #802; otherwise simple regional-baseline query).
  - Summary line: total community observations + new species this week.
[...]

PR #1060, opened 2026-05-15
fix(identify): deploy with --no-verify-jwt so the function self-gates

ArtemioPadilla:
## Summary

`identify` is **built to self-gate auth** but was deployed with `verify_jwt=true`, so Supabase's edge gateway rejected the request before the function ran. Fix mirrors `api`/`mcp`/`share-card`: `verify_jwt=false` in `config.toml` + `identify` added to `OTHER_NO_JWT` in `deploy-functions.yml`.

## Root cause

- The function handles **anonymous callers** (anon IP rate-limit, `index.ts:451-469`) and resolves the JWT **itself** when present (`auth.getUser`, `index.ts:525-534`). Anon + optional-JWT is a designed path.
- But edge `verify_jwt=true` parses the Bearer token as a JWT *before* the function runs. Since the project migrated to `sb_publishable_*` keys, any call sending the publishable key as Bearer — anonymous users, **and** signed-in users whose gotrue session lock stalls (`Lock "lock:rastrum-auth-v1" was not released within 5000ms`) and falls back to the apikey — is rejected at the gateway with `UNAUTHORIZED_INVALID_JWT_FORMAT`.
- Evidence: production `identify` logs show **only** `booted` / `shutdown` — the function body never executes. Reproduced by curling the EF with the publishable anon key → `401 {"code":"UNAUTHORIZED_INVALID_JWT_FORMAT"}`, zero function logs. Symptom to the user: identification silently produces nothing.
- Same failure class as the documented cron pitfall (*"publishable sb_publishable_… key not accepted as Bearer by Edge Functions → deploy --no-verify-jwt"*). `identify` was left out of that list by accident.

## Why this is not a security regression

- Anon callers are already IP rate-limited (`checkAnonRateLimit`, `ANON_LIMIT`/`WINDOW_SEC`).
- Claude still requires a valid JWT to resolve sponsorship/pool — anon cannot spend another user's credit (`beneficiaryId` comes from `auth.getUser`).
- PlantNet operator path is capped by the PlantNet provider quota.
- Identical posture to `api`/`mcp` (token-gated internally, `--no-verify-jwt` at the edge).

## Both changes are required

- `deploy-functions.yml` `OTHER_NO_JWT` — what the deploy **actually applies** (`--no-verify-jwt` flag).
- `config.toml` — documents intent + source of truth for local CLI runs. Without it, intent silently drifts.

## Test plan

- [ ] CI green.
[...]

PR #1069, opened 2026-05-16
fix(deploy): refresh-conservation-status is cron-only — skip JWT verify

ArtemioPadilla:
## Problem

Manually firing `refresh-conservation-status` (after #1068 fixed its unset-GUC error) returned:

```
401 {"code":"UNAUTHORIZED_NO_AUTH_HEADER","message":"Missing authorization header"}
```

That is the Supabase **gateway** rejecting the request (no `Authorization` bearer) before the function code runs — i.e. the EF is deployed `verify_jwt=true`.

`refresh-conservation-status` self-gates on `X-Cron-Secret` (its own `authOk` reading `Deno.env.get('CRON_SECRET')`) and the monthly pg_cron job calls it with no bearer. It was **missing from `CRON_ONLY`** in `deploy-functions.yml` and had no `config.toml` block, so it deployed JWT-verified.

Masked until now: the job built its URL from an unset `current_setting('app.supabase_url')` GUC (#1068), so it errored in Postgres before ever reaching the gateway. Fixing the GUC exposed this — same class as #1066.

## Fix

- Add `refresh-conservation-status` to `CRON_ONLY` (deploy-time `--no-verify-jwt`).
- Mirror `verify_jwt = false` in `config.toml` (both-sites invariant).

## Test plan

- [ ] CI green
- [ ] After merge → deploy-all (WORKFLOW_TOUCHED) → manual fire returns 200, not 401

## Out of scope (separate bug, flagged)

`weekly-digest` now authenticates+runs but 500s with `column users.email does not exist` — its EF selects `email` from `public.users`, which has no such column (email lives in `auth.users`). Feature #868 has never worked. Needs a design decision on email sourcing; not bundled here.

### DIFF
--- a/.github/workflows/deploy-functions.yml
+++ b/.github/workflows/deploy-functions.yml
@@ -211,7 +211,7 @@ jobs:
           # anon key, which Edge Functions reject when JWT verification is
           # enabled. These MUST be deployed with --no-verify-jwt.
           # Update this list when adding new cron-triggered functions.
-          CRON_ONLY="recompute-streaks award-badges gc-orphan-media plantnet-monitor recompute-user-stats streak-push retry-unidentified enrich-taxon refresh-taxon-ranges kairos-fire recompute-taxa-cache weekly-digest"
+          CRON_ONLY="recompute-streaks award-badges gc-orphan-media plantnet-monitor recompute-user-stats streak-push retry-unidentified enrich-taxon refresh-taxon-ranges kairos-fire recompute-taxa-cache weekly-digest refresh-conservation-status"
           echo "cron_only=$CRON_ONLY" >> "$GITHUB_OUTPUT"
 
       - name: Deploy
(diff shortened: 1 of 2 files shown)


Pamela-ruiz9 (commented):
## Review — fix(deploy): refresh-conservation-status is cron-only — skip JWT verify

**CI verde ✅**

Mismo patrón de #1060 / #1066: EF que se auto-gatea con `X-Cron-Secret` pero estaba desplegada con `verify_jwt=true` → gateway 401 antes de que el código corriera. El detalle interesante aquí es que el bug estuvo enmascarado por el error de GUC (#1068) — sin ese fix, nunca habría llegado al gateway para ver el 401.

Los dos cambios son los correctos y mínimos: `CRON_ONLY` en el workflow + bloque `verify_jwt = false` en `config.toml`. Ambos sites en sync, como debe ser.

El follow-up de `weekly-digest` (`column users.email does not exist`) está bien identificado y excluido — es un bug de diseño diferente, no relacionado con JWT.

**LGTM ✅**

